PR #5, opened 2015-10-02
Fix anchor links in main GraphQL docs

saikat:
These are pointing to a page that doesn't seem to exist (api-reference-type-system.html)


saikat:
Actually this is not enough.  It still goes to http://graphql.org/docs/api-reference-graphql/api-reference-type-system#link for the links instead of where it should be going (http://graphql.org/docsl/api-reference-type-system.html#link)

zpao:
I fixed this and a couple others with https://github.com/graphql/graphql.github.io/commit/0bc0f498448f791780eb4a685d858e1665742f74. Thanks for catching the issue!

PR #6, opened 2015-10-08
fix typo

philschatz:
Changed "Developer deal with systems that lack this ..." to "Developers deal with systems that lack this ..."


zpao:
Thanks!

PR #10, opened 2015-10-21
Fix section name for blog layout

craigbilner:
Ensure that when navigating blog articles, blog header link has the active class rather than docs.


zpao:
👍 Thanks!

PR #13, opened 2015-12-10
Add missing `description` field on `GraphQLArgumentConfig` type

albertstill:
See https://github.com/graphql/graphql-js/blob/60e55da31f761303f60bc8258a71d278939e52c0/src/type/definition.js#L491


zpao:
Thanks! (and sorry for the delay)

PR #14, opened 2016-01-27
Update scalar type config to use serialize/parse

Eluinhost:
Also added an extra example that shows how to use string hex external values with integer internal values.


leebyron:
Just landed a similar change, thanks!

PR #16, opened 2016-02-23
Update Scalar doc to match api update

robrichard:


robrichard:
I just noticed that https://github.com/graphql/graphql.github.io/pull/14 is already open to cover this.

PR #20, opened 2016-04-01
Fix walkthrough Queries test file link

hamzahik:


hamzahik:
Just that there is already one pull request fixing this :+1:

PR #24, opened 2016-04-23
Fix a couple typos

lacker:
"Utilities" misspelled in two different ways ;-)


lacker:
Oops, I missed that the only other pull request is exactly the same thing. Just merge that one

PR #27, opened 2016-05-03
Update return types in utilities to match source

alexanderchr:
https://github.com/graphql/graphql-js/blob/master/src/utilities/isValidLiteralValue.js#L38


leebyron:
Thanks!

PR #28, opened 2016-05-05
update graphql function signature to 0.5.0

helfer:
People kept running into problems with this, because documentation wasn't updated.


leebyron:
Whoops! thanks!

PR #29, opened 2016-05-05
Fix (presumably) broken graphql function call

helfer:
I didn't test it, but I think this has probably been broken since 0.5.0 and this commit should fix it.


helfer:
Never mind, this will break the current demo, because that still uses the old graphql-js.

PR #37, opened 2016-08-22
Learn / Queries and Mutations

stubailo:
**Status: draft**

I've covered:

* A simple query
* Nesting
* Arguments
* Aliases
* Fragments
* Comments
* Variables
* Operation names
* Directives (skip & include)
* A simple mutation

### Open questions to resolve

- [x] What do we call these? "Queries" doesn't make total sense because then we're ignoring mutations; "Queries and Mutations" is a bit verbose; "Request" might conflict with HTTP requests.
- [x] Do we want to have this be lots of small pages (one for variables, one for fragments, etc) or one big page? If it's one big page, we'll almost certainly want a ToC either in the sidebar or at the top of the page.
- [x] I'm currently assuming this page isn't the "getting started" introduction - it's a complete guide to different features so that people don't need to read the spec to understand how to use GraphQL. Need to verify this assumption with others.
- [ ] Do we have a schema we want to use for all of the examples? Is it still Star Wars? Going to go with SWAPI for now.
- [ ] I think it would be neat if we just put a `graphql` code snippet, and then some JS code traversed the page later and inserted the interactive GraphiQL thing. Then, you could actually read these on GitHub and the highlighting in my markdown editor wouldn't be messed up. Seems better than an inline `<script>` tag IMO

### DIFF
--- a/site/docs/Learn-Queries.md
+++ b/site/docs/Learn-Queries.md
@@ -1,0 +1,17 @@
+---
+title: Queries and Mutations
+layout: ../_core/DocsLayout
+category: Learn
+permalink: /learn/queries/
+next: /docs/queries/
+---
+
+### Outline
+
+* A simple query
+* Fields, arguments, and aliases
+* Variables
+* Fragments
+* Comments
+* Directives (skip & include)
+* A simple mutation
--- a/site/docs/Learn-Queries.md
+++ b/site/docs/Learn-Queries.md
@@ -6,7 +6,9 @@ permalink: /learn/queries/
 next: /docs/queries/
 ---
 
-### Outline
+### ToC
+
+XXX we should probably generate one
 
 * A simple query
 * Nesting
@@ -61,7 +63,7 @@ In the previous example, we just asked for a simple string, but fields can also 
 `} />);
 </script>
 
-Filtering fields and traversing related objects is pretty nice, but we are just getting started.
+Note that in this example, the `friends` field returns an array of items. In GraphQL, queries don't specify whether the return value should be a single item or an array. You need to know which one to expect ahead of time, but it will always be consistent with what is indicated in the schema.
 
 ### Arguments
 
--- a/site/docs/Learn-Queries.md
+++ b/site/docs/Learn-Queries.md
@@ -9,10 +9,13 @@ next: /docs/queries/
 ### Outline
 
 * A simple query
-* Fields, arguments, and aliases
+* Nesting
+* Arguments
+* Aliases
 * Fragments
 * Comments
 * Variables
+* Operation names
 * Directives (skip & include)
 * A simple mutation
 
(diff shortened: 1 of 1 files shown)


leebyron:
This is awesome, thanks for tackling such a huge spread of content.

stubailo:
@leebyron any thoughts on the questions I posed in the original post?

stubailo:
In particular, the biggest one is if we should define a consistent schema that we use in every example, or if I should just make up queries as I go along.

leebyron:
Here are my thoughts on your open questions:

> What do we call these? "Queries" doesn't make total sense because then we're ignoring mutations; "Queries and Mutations" is a bit verbose; "Request" might conflict with HTTP requests.

I think colloquially we should call them "Queries and Mutations" but we should also formally describe them as "Operations". Such as "Queries and Mutations are two types of Operations - an Operation is a specific action to ask a GraphQL service to perform"

> Do we want to have this be lots of small pages (one for variables, one for fragments, etc) or one big page? If it's one big page, we'll almost certainly want a ToC either in the sidebar or at the top of the page.

Either way we definitely want a ToC sidebar! I think we should have one big page for the scope of content you've covered here, another big page for the type system, but likely individual pages for some of the other more self-contained topics.

> I'm currently assuming this page isn't the "getting started" introduction - it's a complete guide to different features so that people don't need to read the spec to understand how to use GraphQL. Need to verify this assumption with others.

Yeah, we're on the same page. /learn/index.html should be an introduction that touches all relevant topics to arrive at a thin but broad understanding of the whole before diving in here. Your audience framing is also dead on: this is for those who want to understand GraphQL and how to use it, but for whom reading the spec would be overkill or too opaque.

> Do we have a schema we want to use for all of the examples? Is it still Star Wars? Going to go with SWAPI for now.

I feel like we should have consistent schema for examples throughout. SWAPI seems fine to me, though I'm definitely open to better ideas.

> I think it would be neat if we just put a graphql code snippet, and then some JS code traversed the page later and inserted the interactive GraphiQL thing. Then, you could actually read these on GitHub and the highlighting in my markdown editor wouldn't be messed up. Seems better than an inline \<script> tag IMO

That shouldn't be too hard. Maybe we mark them with something like ```graphql!interactive so we have some control over it?

stubailo:
Thanks for the quick reply!

### Action items

- I'll take all of the examples in this article and put together a schema that lets me demonstrate everything I need to. Then we can maybe add to it over time, if we need some examples for union types, interfaces, etc.
- I'll see if it's easy to do the code snippet thing - seems like it should be pretty simple and will make writing this stuff a lot more enjoyable. I agree that we need some way to pass some custom arguments or metadata along with the code snippet.
- After the above, I'll address all of the XXX comments in the article which are blocked on it
- Someone needs to add variable support to the mini GraphiQL - who can do this?

leebyron:
Variable support in mini GraphiQL shouldn't be that hard - I can tackle that.

leebyron:
Also, since this branch doesn't affect anything production right now, I think we should ship-and-iterate. Feel free to land this and then come back to it to amend/add content later if necessary.

stubailo:
Sounds great to me! I hate having long-running branches.

stubailo:
Should I open issues for todo items? Or where should we keep track of outstanding tasks?

leebyron:
Issues works for me.

PR #44, opened 2016-08-31
[WIP] New schema for all query/schema examples

stubailo:
I went on a quest to design a schema that exercises every feature of GraphQL schemas I know of, and this is what I have so far.

The other goal is to reimplement the schema using the schema language directly, rather than having it in comments that are in danger of becoming outdated.

### DIFF
--- a/site/docs/Learn-Schema.md
+++ b/site/docs/Learn-Schema.md
@@ -90,7 +90,7 @@ That means that the GraphQL service needs to have a `Query` type with `hero` and
 ```graphql
 type Query {
   hero(episode: Episode): Character
-  droid(id: String!): Droid
+  droid(id: ID!): Droid
 }
 ```
 
@@ -217,7 +217,7 @@ For example, you could have an interface `Character` that represents any charact
 
 ```graphql
 interface Character {
-  id: String!
+  id: ID!
   name: String!
   friends: [Character]
   appearsIn: [Episode]!
@@ -230,15 +230,16 @@ For example, here are some types that might implement `Character`:
 
 ```graphql
 type Human implements Character {
-  id: String!
+  id: ID!
   name: String!
   friends: [Character]
   appearsIn: [Episode]!
-  homePlanet: String
+  starships: [Starship]
+  totalCredits: Int
 }
 
 type Droid implements Character {
-  id: String!
+  id: ID!
   name: String!
   friends: [Character]
   appearsIn: [Episode]!
@@ -283,17 +284,7 @@ Union types are very similar to interfaces, but they don't get to specify any co
 XXX no example in SWAPI
 
 ```graphql
-union SearchResult = Photo | Person
-
-type Person {
-  name: String
-  age: Int
-}
-
-type Photo {
-  height: Int
-  width: Int
-}
+union SearchResult = Character | Starship
 ```
 
 In this case, if you query a field that returns the `SearchResult` union type, you need to use a conditional fragment to be able to query any fields at all.
--- a/site/docs/Learn-Schema.md
+++ b/site/docs/Learn-Schema.md
@@ -79,7 +79,7 @@ query {
   hero {
     name
   }
-  droid(id: "2001") {
+  droid(id: "2000") {
     name
   }
 }
@@ -170,7 +170,7 @@ The Non-Null type modifier can also be used when defining arguments for a field,
 
 ```graphql
 # { "graphiql": true, "variables": { "id": null } }
-query DroidById($id: String!) {
+query DroidById($id: ID!) {
   droid(id: $id) {
     name
   }
@@ -249,7 +249,9 @@ type Droid implements Character {
 
 You can see that both of these types have all of the fields from the `Character` interface, but also bring in extra fields, `homePlanet` and `primaryFunction`, that are specific to that particular type of character.
 
-Interfaces are useful when you want to return an object or set of objects, but those might be of several different types. For example, in the following query:
+Interfaces are useful when you want to return an object or set of objects, but those might be of several different types.
+
+For example, note that the following query produces an error:
 
 ```graphql
 # { "graphiql": true, "variables": { "ep": "JEDI" } }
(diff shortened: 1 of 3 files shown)


stubailo:
Notes from @tmeasday:

- [ ] Input object types?
- [ ] Mutations not defined
- [ ] Argument to `search` shouldn't be called `query`, it's too vague

site/docs/Learn-Schema.md
```diff
@@ -85,7 +90,7 @@ That means that the GraphQL service needs to have a `Query` type with `hero` and
 ```graphql
 type Query {
   hero(episode: Episode): Character
-  droid(id: ID!): Droid
+  droid(id: String!): Droid
```
leebyron: It's good to encourage the best practice of using ID type for id fields/arguments

site/docs/Learn-Schema.md
```diff
@@ -213,9 +218,9 @@ For example, you could have an interface `Character` that represents any charact
 ```graphql
 interface Character {
   id: String!
```
leebyron: So we should change these to ID instead

site/docs/Learn-Schema.md
```diff
@@ -226,17 +231,17 @@ For example, here are some types that might implement `Character`:
 ```graphql
 type Human implements Character {
   id: String!
-  name: String
+  name: String!
```
leebyron: I think it's probably better to avoid overuse of !. For one, it's noisy when teaching graphql concepts, but also it's a dangerous default state when built on languages that can throw errors

leebyron:
Perhaps for mutations you can add a rating to a film and query ratings? That seems more realistic than adding characters to a film

stubailo:
I was trying to figure out a use case for input objects, and creating a new object seems like the most common use for them. But perhaps we could just skip input objects.

site/docs/Learn-Schema.md
stubailo: Yeah I noticed that the old example schema had String instead of ID, but let's upgrade.

site/docs/Learn-Schema.md
stubailo: On the other hand, in some typed languages having everything be nullable makes it a bit annoying to work with the data. I talked to Dan and he also had a few ideas about it which basically came down to - data that involves fetches should always be nullable in case the fetch fails, but scalar fields on objects that are returned as one unit should maybe be non-null.

But that's mostly an academic concern and I agree too many `!` clutter the schema.

leebyron:
Also while you're at it - take a look at some of the examples on /index.html - they're chosen for simplicity to fit on the screen, but it would be awesome if they aligned well with the schema used throughout examples.

leebyron:
For input objects, how about:

```graphql
input ReviewInput {
  # 0-5 stars
  stars: Int

  # what'd you think?
  commentary: String
}

type Mutation {
  addReview(film: ID, review: ReviewInput): AddReviewResult
}

type AddReviewResult {
  didSucceed: Boolean!
  film: Film
  review: Review
}
```

stubailo:
OK, I see that you're thinking of the schema as strictly about the movies, not as an in-universe thing (adding characters, or giving someone credits, doesn't make sense if you're thinking about it as a movie resource for fans). So the schema use case is "a schema for an app where fans can learn and interact about starwars"

leebyron:
It could go either way, but we need to expand it in some sensible way to support these other featuees for ease of explanation :)

site/docs/Learn-Schema.md
lacker: I don't see why that is a best practice - the only benefit of ID is that you can pass it an int and it will automatically cast to a string, right? normally you would not want to do that in your schema AFAICT. if it's a string in your database it seems safer to have it be a string in graphql.

stubailo: I also don't really see the benefit of IDs, but it seems like a lot of people value them. Dan Schafer went out of his way in his React Europe talk to mention how non-user-readable IDs should be treated specially.

leebyron: Being more generally accepting of a wider range of input is one benefit, but the primary benefit is the semantic value. It distinguishes from strings you might display to a user.

It also helps in migrations when you might start using Int based ids but later realize the scale problem and want to migrate to 64int or String based ids. Or when you need to unify multiple systems which use different storage types for their ids. This is a very real problem we faced at Facebook many times, and so making String/Int backing an implementation detail of the semantic ID offers more freedom for implementors and shields clients from that

leebyron:
Re IDs:

Being more generally accepting of a wider range of input is one benefit, but the primary benefit is the semantic value. It distinguishes from strings you might display to a user.

It also helps in migrations when you might start using Int based ids but later realize the scale problem and want to migrate to 64int or String based ids. Or when you need to unify multiple systems which use different storage types for their ids. This is a very real problem we faced at Facebook many times, and so making String/Int backing an implementation detail of the semantic ID offers more freedom for implementors and shields clients from that

stubailo:
For reference, here are all the code samples from index.html:

```
"hero": {
    "name": "Luke Skywalker",
    "eyeColor": "blue",
    "hairColor": "blond"
  }
```

```
{
  hero {
    name
    friends {
      name
    }
  }
}
```

```
{
  hero {
    name
    friends {
      name
      homeWorld {
        name
        climate
      }
      species {
        name
        lifespan
        origin {
          name
        }
      }
    }
  }
}
```

```
type Query {
  hero: Character
}

type Character {
  name: String
  friends: [Character]
  homeWorld: Planet
  species: Species
}

type Planet {
  name: String
  climate: String
}

type Species {
  name: String
  lifespan: Int
  origin: Planet
}
```

```
type Film {
  title: String
  episode: Int
  releaseDate: String
  openingCrawl: String
  director: String @deprecated
  directedBy: Person
}

type Person {
  name: String
  directed: [Film]
  actedIn: [Film]
}
```

I don't think we should update the example schema to include all of these types, but I think we can rewrite the examples to draw from a more concise schema.

leebyron:
Cool. Doesn't have to be perfect. Front page should optimize for simplicity rather than consistency with the rest of the docs, but places where consistency is a no brainer should change

stubailo:
OK, going to merge this for now since all of the existing examples have been converted.

PR #48, opened 2016-09-07
Add Best Practices: Serving Over HTTP

robzhu:
I'd like to have a section on how to configure a new NodeJS project for use with GraphQL-js/express since I want the other best practices sections to have concrete examples. 

@lacker this might overlap with some of the stuff you're doing for the coding section. 
cc: @stubailo


lacker:
We chatted some over Messenger but for the record, TLDR I don't think using Babel is a best practice with a Node server, I think best practices are to just use Node v6, especially for tutorial code.

stubailo:
For the record, I always use Babel even with Node 6, but who knows what "best practices" even means in the JavaScript ecosystem.

site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -0,0 +1,45 @@
...
+next: /learn/authorization/
+---
+
+HTTP is a great choice for client-server communication when using GraphQL because of its ubiquity. Here are some guidelines for setting up a GraphQL server to operate over HTTP.
+
+### Web Request Pipeline
+Most modern web frameworks use a pipeline model where requests are passed through a stack of middleware (aka filters/plugins). As the request flows through the pipeline, it can be inspected, transformed, modified, or terminated with a response. GraphQL should be placed after all authentication middleware.
+
+### URIs, Routes, and Controllers
+HTTP is commonly associated with REST, which uses "resources" as its core concept. In contrast, GraphQL uses an entity graph as its conceptual model. As a result, entities in GraphQL are not identified by URIs. Instead, a GraphQL server operates on a single URI/endpoint. All GraphQL requests for a given graph/domain should be directed at this endpoint. If you are coming from an MVC framework, think of GraphQL as a unified controller that handles requests for all entities.
```
lacker: I don't get the line about "think of GraphQL as a unified controller" - that doesn't seem very MVC to me.

You might want to say URL because some people still don't know what URI is.

site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -0,0 +1,45 @@
...
+
+### HTTP Methods, Headers, and Body
+Your GraphQL HTTP server should handle the HTTP GET and POST methods.
```
lacker: This isn't a formal part of the spec, right? It's just how we do it. So I think this part should be "suggested" or something rather than "should". IMO it would be fine to write your own graphql server that only handled POST, if you wanted to.

site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -0,0 +1,45 @@
...
+
+When receiving an HTTP GET request, the GraphQL query should be specified in the "query" query string. For example:
+
+```
+curl "http://myapi/graphql?query={me{name}}"
+```
+*Note: The '{' and '}' characters need to be URL encoded, but they are intact above for clarity*
```
lacker: Having an example that doesn't work is not good for clarity

site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -0,0 +1,45 @@
...
+
+When receiving an HTTP POST request, we recommend supporting these two cases:
+
+* If the "query" query string parameter is present (as in the GET example above), it should be parsed and handled in the same way as the HTTP GET case.
+* If the "application/graphql" Content-Type header is present, treat the HTTP POST body contents as the GraphQL query string.
+
+If you're using express-graphql, you already get these behaviors for free.
+
+### GraphiQL
+GraphiQL is useful during testing and development but disabled in production. If you are using express-graphql, you can toggle it based on the NODE_ENV environment variable:
```
lacker: In some cases it seems like you'd want to use it in production, so I'd hedge this a bit

site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -0,0 +1,45 @@
...
+
+```
+app.use('/graphql', graphqlHTTP({
+  schema: MySessionAwareGraphQLSchema,
+  graphiql: process.env.NODE_ENV === 'development',
+}));
+```
+
+### Node
+If you are using NodeJS, we recommend using either express-graphql or apollo-server.
```
lacker: Linkify these?

site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Serving over HTTP
+layout: ../_core/DocsLayout
+category: Best Practices
+permalink: /learn/serving-over-http/
+next: /learn/authorization/
+---
+
+HTTP is a great choice for client-server communication when using GraphQL because of its ubiquity. Here are some guidelines for setting up a GraphQL server to operate over HTTP.
```
lacker: Could be a bit more specific and say it's the best / most common choice. It seems unlikely you'd want something other than HTTP, unless you want something more realtimey.

robzhu:
Diff is really noisy from merging too late, opening a new PR.

PR #49, opened 2016-09-08
Code page and GraphQL.js docs

lacker:
This supports both the /code/ page and a bunch of graphql.js-specific docs that start with /graphql-js/getting-started/.

The sidebar doesn't work yet, but I wanted to get this in so that we can link to it if we want to with things like https://github.com/graphql/graphql.github.io/pull/48 . cc @robzhu


site/code/index.html.js
```diff
@@ -17,41 +17,36 @@ var Code = React.createClass({
...
+  - [graphql-ruby](https://github.com/rmosolgo/graphql-ruby): A Ruby library for building GraphQL APIs. Built-in support for Relay and Rails.
+  - [graphql-java](https://github.com/graphql-java/graphql-java): A Java library for building GraphQL APIs.
+  - [graphql-relay](https://github.com/graphql/graphql-relay-js) ([npm](https://www.npmjs.com/package/graphql-relay)): A server library to help construct a Node.js GraphQL API that supports [Relay](https://facebook.github.io/relay/).
+  - [Sangria](http://sangria-graphql.org/) ([github](https://github.com/sangria-graphql/sangria)) A Scala GraphQL library that supports [Relay](https://facebook.github.io/relay/).
+  - [Apollo Server](http://dev.apollodata.com/tools/apollo-server/index.html) ([github](https://github.com/apollostack/apollo-server)) ([npm](https://www.npmjs.com/package/apollo-server)): A GraphQL server that works with all Node.js HTTP server frameworks: Express, Connect, HAPI and Koa. Designed to work well with the Apollo Client.
```
stubailo: Apollo Server doesn't have any connection to Apollo Client - so let's remove that part

site/code/index.html.js
```diff
@@ -17,41 +17,36 @@ var Code = React.createClass({
...
 
-  - **[graphql](https://github.com/facebook/graphql)** ([html](http://facebook.github.io/graphql/)): A working draft of the specification for GraphQL.
+## GraphQL Clients
 
-### Server (JavaScript)
+  - [Relay](https://facebook.github.io/relay/) ([github](https://github.com/facebook/relay)) ([npm](https://www.npmjs.com/package/react-relay)): Facebook's framework for building React applications that talk to a GraphQL backend.
+  - [Apollo Client](http://dev.apollodata.com/react/) ([github](https://github.com/apollostack/apollo-client)): A powerful JavaScript GraphQL client, designed to work well with the Apollo Server.
```
stubailo: Apollo client works equally well with every GraphQL server, and doesn't require any special features at all. So let's remove that. I'd also emphasize that it works with React and Angular 2.

site/docs/Guides-ConstructingTypes.md
```diff
@@ -0,0 +1,116 @@
...
+layout: ../_core/DocsLayout
+category: Guides
+permalink: /graphql-js/constructing-types/
+---
+
+For many apps, you can define a fixed schema when the application starts, and define it using GraphQL schema language. In some cases, it's useful to construct a schema programmatically. You can do this using the `GraphQLSchema` constructor.
+
+When you are using the `GraphQLSchema` constructor to create a schema, instead of defining `Query` and `Mutation` types solely using schema language, you create them as separate object types using the `GraphQLObjectType` constructor.
+
+For example, let's say we are building a simple API that lets you fetch user data for a few hardcoded users based on username. Using `buildSchema` we could write a server with:
```
robzhu: Your example seems to fetch by user id rather than username.

site/docs/Tutorial-BasicTypes.md
```diff
@@ -0,0 +1,58 @@
...
+title: Basic Types
+layout: ../_core/DocsLayout
+category: Tutorial
+permalink: /graphql-js/basic-types/
+next: /graphql-js/passing-arguments/
+---
+
+In most situations, all you need to specify the types for your API is the GraphQL schema language, taken as an argument to the `buildSchema` function.
+
+The GraphQL schema language supports the scalar types of `String`, `Int`, `Float`, `Boolean`, and `ID`, so you can use these directly in the schema you pass to `buildSchema`.
```
robzhu: It might be worth mentioning GraphQLEnumType here.

site/docs/Tutorial-ExpressGraphQL.md
```diff
@@ -0,0 +1,58 @@
...
+console.log('Running a GraphQL API server at localhost:4000/graphql');
+```
+
+You can run this GraphQL server with:
+
+```bash
+node server.js
+```
+
+Since we configured `graphqlHTTP` with `graphiql: true`, you can use the GraphiQL tool to manually issue GraphQL queries. If you navigate in a web browser to `http://localhost:4000/graphql`, you should see an interface that lets you enter queries. It should look like:
```
robzhu: It's really easy to forget to disable GraphiQL for production. I suggest: 

graphiql: process.env.NODE_ENV !== 'production'

site/docs/Tutorial-GraphQLClients.md
```diff
@@ -0,0 +1,34 @@
+---
+title: GraphQL Clients
+layout: ../_core/DocsLayout
+category: Tutorial
+permalink: /graphql-js/graphql-clients/
+next: /graphql-js/basic-types/
+---
+
+Since a GraphQL API has more underlying structure than a REST API, there are more powerful clients like [Relay](https://facebook.github.io/relay/) which can automatically handling batching, caching, and other features. But you don't need a complex client to call a GraphQL server. With `express-graphql`, you can just send an HTTP POST request to the endpoint you mounted your GraphQL server on, passing the GraphQL query as the `query` field in a JSON payload.
```
robzhu: Can we add a curl example here? Also, I would mention that queries can be executed with an HTTP GET, but mutations must be executed with an HTTP POST.

site/docs/Tutorial-ObjectTypes.md
```diff
@@ -0,0 +1,144 @@
...
+  getDie(numSides: 6) {
+    rollOnce
+    roll(numRolls: 3)
+  }
+}
+```
+
+If you run this code with `node server.js` and browse to [http://localhost:4000/graphql](http://localhost:4000/graphql) you can try out these APIs with GraphiQL.
+
+This way of defining object types often provides advantages over a traditional REST API. Instead of doing one API request to get basic information about an object, and then multiple subsequent API requests to find out more information about that object, you can get all of that information in one API request. That saves bandwidth and makes your app run faster.
```
robzhu: In addition to perf, I would mention that the single request/response model simplifies client-side async logic, error handling, sequencing, and de-duping.

site/code/index.html.js
lacker: OK

site/code/index.html.js
lacker: OK

site/docs/Guides-ConstructingTypes.md
lacker: fixed

site/docs/Tutorial-ExpressGraphQL.md
lacker: Hmm, that seems like a reasonable practice but I don't really want to make all of the example code include the extra concept of "what is NODE_ENV" because I think many people won't understand that or use that convention.

site/docs/Guides-ConstructingTypes.md
```diff
@@ -0,0 +1,116 @@
...
+    id: 'a',
+    name: 'alice',
+  },
+  'b': {
+    id: 'b',
+    name: 'bob',
+  },
+};
+
+var root = {
```
stubailo: I think it's a bit confusing that the example that uses the schema language does everything via the root value, and the one with GraphQLObjectType uses resolvers. Right now this looks like "here's a simple way to do it and a complicated way", but in reality they are just two different ways to construct a server, with different tradeoffs.

I think something like `graphql-tools` makes this distinction smaller because it lets you write resolvers and also use the schema language, but I think resolvers are a core part of GraphQL that probably shouldn't be papered over in the introduction.

site/docs/Tutorial-BasicTypes.md
stubailo: We talk about all of these features of the schema language in the Learn section, probably worth linking to that.

site/docs/Tutorial-GraphQLClients.md
```diff
@@ -0,0 +1,34 @@
...
+});
+```
+
+You should see the data returned, logged in the console:
+
+```
+data returned: Object { hello: "Hello world!" }
+```
+
+The `fetch` API used here is in most modern browsers - check out [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) if you'd like to learn more. Any mechanism of sending HTTP requests should work similarly. In the browser, we recommend using `fetch` over `XMLHttpRequest` if you have the choice because the API is nicer. In a mobile application, you can use whatever library you'd normally use to send HTTP requests.
```
stubailo: It's not in Safari... here's what I needed to do to run a GraphQL query from "plain JavaScript" https://github.com/stubailo/microhn/blob/gh-pages/index.html#L92

site/docs/Tutorial-GraphQLClients.md
stubailo: It would be nice to link to the "code" section that lists a variety of clients here, rather than Relay directly.

site/docs/Tutorial-BasicTypes.md
lacker: You have to define enums with the 'enum' keyword so I think they belong in some other doc, after the "basic types" section. So let me punt on that til later.

I should confess, I have never used an enum type in graphql, I just pass strings :P

site/docs/Tutorial-GraphQLClients.md
lacker: I added a curl example. There are several ways to send queries - like you can set content type to application/graphql. I'd rather keep this tutorial simple and just explain one way to do things - I would link to the reference docs for express-graphql, but they don't really exist for this yet. So let's only mention the POST mechanism & put the extensive explanation in reference docs for the 'graphqlHTTP' method later.

site/docs/Guides-ConstructingTypes.md
lacker: We chatted about this over the phone but TLDR this example comes later under "guides" so I think the ordering is clear

site/docs/Guides-ConstructingTypes.md
```diff
@@ -0,0 +1,116 @@
...
+var app = express();
+app.use('/graphql', graphqlHTTP({
+  schema: schema,
+  graphiql: true,
+}));
+app.listen(4000);
+console.log('Running a GraphQL API server at localhost:4000/graphql');
+```
+
+When we use this method of creating the API, the root level resolvers are implemented on the `Query` and `Mutation` types rather than on a `root` object.
```
hramos: Where are Mutations covered in this example?

site/docs/Tutorial-BasicTypes.md
stubailo: We have it in this section: https://github.com/graphql/graphql.github.io/blob/30c5738b927505f162ce585d26727b9630d63b4f/site/docs/Learn-Schema.md#enumeration-types

site/docs/Tutorial-GraphQLClients.md
lacker: You're right - fetch kinda sucks! TIL about xhr.onload, I had been using onreadystatechange and checking for state 4 forever. I changed this example to just use xhr's - it's like 9 lines of code instead of 8, definitely better if you can avoid teaching people about what is fetch and what are promises.

site/docs/Tutorial-ExpressGraphQL.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Running an Express GraphQL Server
+layout: ../_core/DocsLayout
+category: Tutorial
+permalink: /graphql-js/running-an-express-graphql-server/
+next: /graphql-js/graphql-clients/
+---
+
+The simplest way to run an GraphQL API server is to use Express, a popular web application framework for Node.js. (link to https://expressjs.com/). You will need to install two additional dependencies:
```
hramos: Linkify Express here.

site/docs/Tutorial-ObjectTypes.md
lacker: good point OK updated this sentence a bit

site/docs/Tutorial-ExpressGraphQL.md
lacker: Oops! Fixed this ;-)

site/docs/Guides-ConstructingTypes.md
lacker: It isn't in the code example - I basically just used the code example that we currently have in graphql.org so that we don't lose out on any currently-existing data. I wanted to mention it briefly here though... I think this is ok?

site/docs/Tutorial-BasicTypes.md
lacker: Yeah, for using them in actual code you have more stuff to explain though. Like at first I assumed enums mapped to 0, 1, 2 etc because they do in languages like Go and I wondered wasn't that a pain on the client-side. Enums are cool, I just don't think conceptually they are as basic as string / int / bool / float

lacker:
OK I think I have responded to all feedback and accepted most of it ;-) I can still take more feedback but I'd like to merge this in so that I can work on the side bar stuff. Good to merge in - @robzhu ?

robzhu:
Merging for now so we can keep moving. We'll make more edits as more feedback comes in.

PR #52, opened 2016-09-09
Replace a bunch of no-longer-existing pages with redirects

lacker:
Saves a bit of SEO juice & slightly improves the whole experience. Annoyingly our static site generation scripting doesn't have anything built in for redirects, so, there's just a bunch of html redirect files.

@robzhu


lacker:
FYI I accidentally also added code to alter page titles to this PR.

PR #58, opened 2016-09-11
Added `operationName`

OlegIlyenko:


site/docs/BestPractice-ServingOverHTTP.md
```diff
@@ -44,7 +44,8 @@ A standard GraphQL POST request should use the `application/json` content type,
 ```js
 {
   "query": "...",
-  "variables": { variable1: value, ... }
+  "oprationName": "...",
```
stubailo: I think you're missing the `e` in `operationName` here and also above.

Also, I think it's worth mentioning that the `operationName` in the POST body is optional, and only required if you have multiple operations?

As an additional wrinkle, express-graphql doesn't support multiple operations in one document as far as I know.

OlegIlyenko: OMG, sorry for the typo :flushed:

I was also thinking about adding a note about optional fields. In this case we need to mention that `variables` is optional as well. WDYT?

stubailo: Yeah, maybe just add a sentence underneath the code snippet saying operationName and variables are optional.

OlegIlyenko: Perfect. I already added it and rebased the PR

stubailo:
Just one last question - should we be worried that `express-graphql` doesn't support this feature?

OlegIlyenko:
Are you sure about this limitation? I just double checked the docs and looks like `operationName` is supported: https://github.com/graphql/express-graphql#http-usage

stubailo:
Oh, OK I'm totally wrong :]

PR #64, opened 2016-09-12
Add Best Practices: Thinking In Graphs

robzhu:
This should be the first link in the best practices section. Please take a look: @leebyron, @lacker, @stubailo


stubailo:
Looks great to me!

site/docs/BestPractice-ThinkingInGraphs.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+### Business Logic Layer
+> Your business logic layer should act as the single source of truth for enforcing business domain rules
+
+Where should we define the actual business logic? Where should we perform validation and authorization checks? The answer: inside a dedicated business logic layer. Your business logic layer should act as the single source of truth for enforcing business domain rules.
+
+![Business Logic Layer Diagram](/img/diagrams/business_layer.png)
+
+In the diagram above, all entrypoints (REST, GraphQL, and RPC) into the system will be processed with the same validation, authorization, and error handling rules.
```
lacker: entrypoints -> entry points

site/docs/BestPractice-ThinkingInGraphs.md
```diff
@@ -0,0 +1,65 @@
...
+
+Before defining your schema, examine the everyday language you use to describe your business domain rules and processes. For example, let's try to describe an email app in plain english:
+
+* A user can have multiple email accounts
+* Each email account has an address, inbox, drafts, deleted items, and sent items
+* Each email has a sender, receive date, subject, and body
+* Senders have an email address
+* Users cannot send an email without a recipient address
+
+Naming things is a hard but important part of building intuitive APIs, so take time to carefully think about what makes sense for your problem domain and users. Your team should develop a shared understanding and consensus of these business domain rules because you will need to choose intuitive, durable names for nodes and relationships in the GraphQL schema. Try to imagine some of the queries you will want to execute:
```
lacker: I think in most cases, programmers wouldn't think of these as "nodes in a graph" - they would think of them as objects in object-oriented programming, or entries in a database. Do you want to mention something around that? In particular, I have met a lot of people who think graphql is for graph databases. It isn't - it's really nice for both regular databases, and regular object-oriented programming. Those are the two concepts that I think most people will be mapping GraphQL to, here.

site/docs/BestPractice-ThinkingInGraphs.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Thinking in Graphs
+layout: ../_core/DocsLayout
+category: Best Practices
+permalink: /learn/thinking-in-graphs/
+---
```
lacker: I think you'll want a "next" link here

site/docs/BestPractice-ThinkingInGraphs.md
robzhu: Thanks, I added a mention of OOP and did some refactoring on that section.

PR #65, opened 2016-09-12
Best Practices: Authorization

robzhu:
Last Best Practices section for launch. Specifically, I'd like to hear some suggestions on how to make this content actionable without relying so heavily on GraphQL.js. @stubailo @lacker

### DIFF
--- a/site/learn/BestPractice-Authorization.md
+++ b/site/learn/BestPractice-Authorization.md
@@ -6,7 +6,9 @@ permalink: /learn/authorization/
 next: /learn/serving-over-http/
 ---
 
-Authorization rules are a type of business logic that describe whether a given user/session/context has permission to perform a given action or see a piece of data. For example:
+> Delegate authorization logic to the business logic layer
+
+Authorization is a type of business logic that describes whether a given user/session/context has permission to perform an action or see a piece of data. For example:
 
 *“Only authors can see their drafts”*
 
@@ -30,9 +32,9 @@ var postType = new GraphQLObjectType({
 });
 ```
 
-Notice that we define “author owns a post" by checking whether the post's authorId field equals the current user’s id. This is an implementation detail that should be encapsulated in the business domain layer.
+Notice that we define “author owns a post" by checking whether the post's `authorId` field equals the current user’s `id`. Can you spot the problem? We would need to duplicate this code for each entry point into the service. Then if the authorization logic is not kept perfectly in sync, users could see different data depending on which API they use. Yikes! We can avoid that by having a [single source of truth](/learn/thinking-in-graphs/#business-logic-layer) for authorization.
 
-Defining authorization logic inside the resolver is fine when learning GraphQL or prototyping. For production, we recommend delegating this kind of decision to the business logic layer. Here’s what we mean:
+Defining authorization logic inside the resolver is fine when learning GraphQL or prototyping. However, for a production codebase, delegate authorization logic to the business logic layer. Here’s an example:
 
 ```javascript
 //Authorization logic lives inside postRepository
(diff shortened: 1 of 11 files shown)


stubailo:
I think this is fine, because even though the example code is using graphql.js, the text can be applied to any language.

site/learn/BestPractice-Authorization.md
```diff
@@ -3,6 +3,56 @@ title: Authorization
...
+
+```javascript
+var postType = new GraphQLObjectType({
+  name: ‘Post’,
+  fields: {
+    body: {
+      type: GraphQLString,
+      resolve: function (post, args, context, rootValue) {
+        // return the post body only if the user is the post's author
+        if(context.user && (context.user.id === post.authorId)) {
```
lacker: space after 'if'

site/learn/BestPractice-Authorization.md
```diff
@@ -3,6 +3,56 @@ title: Authorization
...
+
+Enforcing this kind of behavior should happen in the [business logic layer](/learn/thinking-in-graphs/#business-logic-layer). It is tempting to place authorization logic in the GraphQL layer like so:
+
+```javascript
+var postType = new GraphQLObjectType({
+  name: ‘Post’,
+  fields: {
+    body: {
+      type: GraphQLString,
+      resolve: function (post, args, context, rootValue) {
```
lacker: do you put a space after "function"? I thought generally no but not sure your/our style guide here

site/learn/BestPractice-Authorization.md
```diff
@@ -3,6 +3,56 @@ title: Authorization
...
+        return postRepository.getBody(context.user, post);
+      }
+    }
+  }
+});
+```
+
+In the example above, we see that the business logic layer requires the caller to provide a user object. If you are using GraphQL.js, the User object should be populated on the `context` or `rootValue` arguments of the resolver.
+
+We recommend passing a fully-hydrated User object instead of an opaque token or API key your business logic layer. This way, we can handle the distinct concerns of [authentication](/graphql-js/authentication-and-express-middleware/) and authorization in different stages of the request processing pipeline.
```
lacker: This sentence seems like it's missing a preposition. "to your business logic layer" ?

lacker:
Do you also need to change the "next" link in Learn-Schema.md ? In general just check the sidebar briefly to make sure it looks good.

site/learn/BestPractice-Authorization.md
stubailo: Generally people do put a space after all keywords - `if`, `function`, etc. But I'd maybe use an arrow function here.

lacker: wow, i checked the airbnb js style guide and they agree with you. turns out i'm the crazy one

lacker: On the arrow functions - everyone should know the function () {} syntax, not all people know the () => {} syntax. so when you aren't using 'this', for doc-writing I think it's more approachable to use the older function () {} style. Even though if I were writing real code in a real codebase I'd prefer () => {}

stubailo: I dunno, basically all React and GraphQL code samples I've seen use the arrow function style, and in particular `() => ({ ... })`, a ton.

lacker: Many popular projects are still documented without arrow functions. Express is a good example - https://expressjs.com/en/guide/routing.html So is Lodash - https://lodash.com/docs/4.15.0#flatMap - I agree that most GraphQL docs nowadays tend to assume you already know how to set up transpilation, but I think that is actually slowing adoption. Promises are the only case where I would suggest using arrow functions, because they are just too verbose with the `function` keyword

robzhu: Thanks, I edited to make this consistent with the GraphQL.js sections.

PR #66, opened 2016-09-12
Misc small docs fixes

lacker:
* Add a small mention of express-graphql in the API reference
* Add API reference for the buildSchema function
* Use `function () {}` style instead of `function() {}` style for consistency with learn docs
* Make express-graphql link on code page link to better place
* Give examples of how to use the `variables` arg in the build-your-own-client example code


site/docs/Tutorial-Mutations.md
```diff
@@ -87,16 +87,16 @@ var fakeDatabase = {};
 
 // The root provides the top-level API endpoints
 var root = {
-  getMessage: function({author}) {
+  getMessage: function ({author}) {
```
leebyron: maybe this style should be:

```js
getMessage({ author }) {
  ...
}
```

instead of the `: function`

leebyron:
lgtm merge at your whim

site/docs/Tutorial-Mutations.md
lacker: That's how I would write it in my own code, but for the tutorial I think it is nice to use minimal fancy ecmascript features so that these docs are approachable to as many people as possible.

lacker: In particular in https://github.com/lacker/graphql.github.io/blob/2c1bb0d86257d7fc66c86d5a4a85064b7717beaf/site/docs/Tutorial-GettingStarted.md we say, hey you might want to read up on es6 classes, promises, and fat arrow functions. I think those are the only es6 features in all this, plus the fact that backticks let you do multiline strings, which is kind of obvious what it does from example code.

PR #71, opened 2016-09-14
SEO the graphql.js part

lacker:
Thought I did this before but probably got lost in the shuffle. Basically this just adds something like "GraphQL.js API Reference" to the title on pages in that category.


lacker:
Somehow this PR is totally merge-borked - i'm gonna open a new one

PR #74, opened 2016-09-14
[bestpractices] Add friends help and cursors to page info

dschafer:
Address feedback in #69, including a start and end cursor in page info, and mentioning the possibility of a convenience method on connections to avoid needing edges.


stubailo:
This is great! Love this simpler way to use connections.

dschafer:
> This is great! Love this simpler way to use connections.

Sweet! Merging now.

PR #82, opened 2016-09-15
Scala example

OlegIlyenko:
I saw a bunch of hello world examples for other languages. I just thought that it would be nice add one for scala as well :)


leebyron:
Sweet! thanks!

lacker:
Thanks man... I just haven't written Scala before so I was having trouble figuring it out ;-) Do you have a "hello world" example somewhere on your website already? I couldn't find one. If not, it might be handy

PR #85, opened 2016-09-15
fix missing return types

lacker:
When I changed the mutations example to use variables correctly, I broke the return types. This fixes the example to correctly have return types for each endpoint & have separate Message and MessageInput types.

### DIFF
--- a/site/graphql-js/Tutorial-Mutations.md
+++ b/site/graphql-js/Tutorial-Mutations.md
@@ -8,17 +8,19 @@ next: /graphql-js/authentication-and-express-middleware/
 
 If you have an API endpoint that alters data, like inserting data into a database or altering data already in a database, you should make this endpoint a `Mutation` rather than a `Query`. This is a simple as making the API endpoint part of the top-level `Mutation` type instead of the top-level `Query` type.
 
-Let's say we have a “message of the day” server, where anyone can update the message of the day along with the author of the message, and anyone can read the current one. The GraphQL schema for this is simply:
+Let's say we have a “message of the day” server, where anyone can update the message of the day, and anyone can read the current one. The GraphQL schema for this is simply:
 
 ```javascript
 type Mutation {
-  setMessage(message: String)
+  setMessage(message: String): Boolean
 }
 
 type Query {
   getMessage: String
 }
 ```
+
+Since `setMessage` doesn't need to return anything useful, we can just return a `Boolean`.
 
 Both mutations and queries can be handled by root resolvers, so the root that implements this schema can simply be:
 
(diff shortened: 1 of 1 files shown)


site/graphql-js/Tutorial-Mutations.md
```diff
@@ -8,25 +8,28 @@ next: /graphql-js/authentication-and-express-middleware/
 
 If you have an API endpoint that alters data, like inserting data into a database or altering data already in a database, you should make this endpoint a `Mutation` rather than a `Query`. This is a simple as making the API endpoint part of the top-level `Mutation` type instead of the top-level `Query` type.
 
-Let's say we have a “message of the day” server, where anyone can update the message of the day along with the author of the message, and anyone can read the current one. The GraphQL schema for this is simply:
+Let's say we have a “message of the day” server, where anyone can update the message of the day, and anyone can read the current one. The GraphQL schema for this is simply:
 
 ```javascript
 type Mutation {
-  setMessage(message: String)
+  setMessage(message: String): Boolean
```
leebyron: More typically graphql mutations return the thing they were mutating so clients can be aware of anything the server did (maybe you lost a race condition with set-message with someone else, or maybe the server did some post-processing) but also as a convenient mechanism to keep a client cache up to date.

That's perhaps overkill here, and illustrating that it's okay to return a boolean is good, but something like this should probably be mentioned here in reference to the fuller example you have below

site/graphql-js/Tutorial-Mutations.md
```diff
@@ -36,25 +39,33 @@ var root = {
 
 You don't need anything more than this to implement mutations. But in many cases, you will find a number of different mutations that all accept the same input parameters. A common example is that creating an object in a database and updating an object in a database often take the same parameters. To make your schema simpler, you can use “input types” for this, by using the `input` keyword instead of the `type` keyword.
 
-For example, instead of a single message of the day, let's say we have one message per author, where the author is just a unique username, and we want a mutation API both for creating a new message and for updating an old message. We could use the schema:
+For example, instead of a single message of the day, let's say we have many messages, indexed in a database by an `ID`, and each message has both a `content` string and an `author` string. We want a mutation API both for creating a new message and for updating an old message. We could use the schema:
```
leebyron: lowercase `id` to match the field name you're referencing?

site/graphql-js/Tutorial-Mutations.md
lacker: OK i changed the example and mentioned this a bit in the text

site/graphql-js/Tutorial-Mutations.md
lacker: ok

PR #99, opened 2016-09-15
Adding our Dreamforce talk with @pmdartus

rricard:
This is about GraphQL in a quite different context: Salesforce! We are glad however to have put this talk together and introduce it to people that may have not been exposed to GraphQL before.


leebyron:
Awesome! Good luck with the talk!

PR #115, opened 2016-09-16
Moved scala.io talk to the right section

OlegIlyenko:
Sorry, I have placed the talk in the "Events" section, which is the wrong place. This PR fixes it.


stubailo:
Good call!

PR #118, opened 2016-09-17
Added GraphQL's Facebook Group to Community section

brunolemos:
Disclaimer: I've created it in March this year.


stubailo:
CC @theadactyl

leebyron:
Oh! I didn't even know this existing, I was actually just about to create one.

Would you mind adding me as an admin first? If we're going to link to a public Facebook group on the site, I'd like to make sure the team is in there and that we're using the standard GraphQL iconography.

brunolemos:
@leebyron No problem! Done.

Issue #133, opened 2016-09-22
Cant access `graphql.org/docs/introspection`


lpalmes:
It seems that it escapes the js script when trying to access [introspection](http://graphql.org/docs/introspection/)

It respond with this piece as html escaped text
```
var React = require('react') var Redirect = require('../../_core/Redirect') export default () =>
```

j0k3r:
Because there is nothing more :trollface: 
https://github.com/graphql/graphql.github.io/blob/source/site/docs/introspection/index.html

It should be a `.html.js`

PR #134, opened 2016-09-22
Fix doc/introspection redirection

j0k3r:
Should fix https://github.com/graphql/graphql.github.io/issues/133


stubailo:
Good call, thanks!

PR #153, opened 2016-10-01
Correction in documentation about importing express-graphql

kalinchernev:
Unifying imported modules.


stubailo:
Good call, thanks!

PR #157, opened 2016-10-02
Update Tutorial-PassingArguments.md

iwaldman:
Use ES6 let and const instead of var.  This is best practice and propose that GraphQL docs move in this direction.

If you agree, happy to review remainder of doc and make similar edits.


stubailo:
This wasn't done accidentally, the idea was to write the tutorial in a way that can run out of the box on Node 4 I think. Cc @lacked

iwaldman:
I suppose since Node 4 is currently LTS this makes sense. Thanks.

PR #158, opened 2016-10-04
Fix underscores being parsed in introspection doc

astorije:
Escaping underscores using backslashes to avoid Markdown treating them
as bold and missing the last one.

### Results (via Atom editor though)

Before | After
--- | ---
screen shot 2016-10-03 at 20 17 44 | screen shot 2016-10-03 at 20 18 08


stubailo:
Good idea, thanks for the screenshots!

astorije:
My pleasure, thanks for the awesomely done website!!

PR #160, opened 2016-10-06
Fix: RootValue is a subvalue of fourth argument in the resolver method

nodkz:


stubailo:
I feel like we should just get rid of `rootValue` entirely - isn't `context` _always_ the right way to do this?

nodkz:
Eg. DB connection pool is host specific (not request/session) so it better to keep in `rootValue`.

Conversation in past https://github.com/nodkz/react-relay-network-layer/issues/9#issuecomment-234465044

stubailo:
> DB connection pool is host specific (not request/session) so it better to keep in rootValue.

Why does this mean it can't be part of context?

stubailo:
> Based on my understanding, context is for session based data, such as current user, etc, while rootValue is for "everything else."

I disagree with this assessment - I think `rootValue` should almost never be used, and if it's used then only for actual data, not configuration. All configuration-type things should like on `context` IMO.

nodkz:
I don't mind. All can be placed to `context`. Without `rootValue` examples more simple, but hide additional possibility to pass data to resolver.

Suggest to wait for the opinion of others.

stubailo:
For now I'll merge this, since the conversation can be had asynchronously.

PR #161, opened 2016-10-06
Add Wayfair logo

adambaratz:
Wayfair is one of the world's largest online destinations for the home. We use GraphQL on our customer-facing storefront and with partner APIs.


stubailo:
Thanks a lot!

PR #164, opened 2016-10-07
Fix .buttons-unit width. Align buttons to center

dru:
screen shot 2016-10-07 at 16 24 48


stubailo:
That's odd - how do I reproduce the buttons being misaligned? They seem to always be centered on my screen.

dru:
It is misaligned under Safari 10 on macOS Sierra

screen shot 2016-10-12 at 1 56 42 pm

stubailo:
Ah, that's right. Thanks!

PR #169, opened 2016-10-13
Add BlenderBottle to the Users page.

alexanderson1993:


alexanderson1993:
BlenderBottle is a consumer products company based in Utah, USA. We use GraphQL in a Shipping Container Manager application used to track shipments from overseas vendors. Plans are in place to expand use to additional applications in both BlenderBottle's internal intranet as well as customer-facing applications.

stubailo:
That's awesome, thanks for adding the logo!

PR #183, opened 2016-10-27
`whitescape.com` added to `/users`

romanonthego:
hey!
since we have used graphql on lmbd.ru (soon to be added to [cases](http://whitescape.com/cases)) we decided it would be nice to have our logo here :)


leebyron:
Great!

PR #191, opened 2016-10-28
Correct `ofType` name for`id` to be `ID`

eventualbuddha:
I'm not sure I understand, but in the example after this they do include`ofType` and end up with this:

``` json
{
          "name": "id",
          "type": {
            "name": null,
            "kind": "NON_NULL",
            "ofType": {
              "name": "ID",
              "kind": "SCALAR"
            }
          }
        },
```

Note that not the name nor the kind is`String`.


stubailo:
Thanks!

PR #192, opened 2016-10-28
add WaitlessQ logo to users

yanivnizry:


leebyron:
Could you provide a higher resolution image? Most images are around the ~400px in the larger dimension.

yanivnizry:
is this enough

PR #193, opened 2016-10-28
Add Artsy to users.

alloy:
[We’re using GraphQL](https://github.com/artsy/metaphysics) to power our [web app](https://github.com/artsy/force) and [iOS app](https://github.com/artsy/emission).


stubailo:
Thanks!

PR #194, opened 2016-10-31
Adding Scaphold

vning93:
Scaphold.io: GraphQL Backend as a Service


stubailo:
@vning93 can you separately PR for the logo in the users page and the community resources page, please?

vning93:
@stubailo I actually tried doing that to begin with but git automatically squashed my commits into the same PR. Is that alright?

vning93:
Just updated given your feedback. Thanks for putting us on the site!

asiandrummer:
@vning93 - sorry to take so long to respond.
> I actually tried doing that to begin with but git automatically squashed my commits into the same PR. Is that alright?

Not trying to be granular, but I also think it'd be better to have two different PRs that touch different parts of the pages. I'd appreciate if you could submit another PR for the community resources page please :D

I think we're ready to merge the logo, but I'm seeing a merge conflict. Could you also resolve that?

PR #195, opened 2016-10-31
Fix field names in Learn Schema example of Character

garno:
Fix a small error in the documentation mentioning field `homePlanet` that is not appearing anywhere else.

✌️


leebyron:
Sweet, thanks for the careful read!

PR #196, opened 2016-10-31
Add M1 Finance to the 'who's using' page.

ryanspradlin:
[Our web, Android, and iOS apps](https://www.m1finance.com/) are powered by a GraphQL API.


leebyron:
Nice!

PR #199, opened 2016-11-03
Adjust homepage media queries to works correctly on mobile

garno:
I wanted to continue reading the documentation on the go and noticed that the homepage wasn't as mobile friendly as it could have been. There was an issue making that made the homepage scroll horizontally and the text was really small. I took a little time to fix it and propose the updated styles.

There is a remaining issue with the _Evolve your API without versions_ example section. I wanted to know if I wasn't wasting my time before trying to fix it :) The issue was already present. If this PR is something you might want to merge, I'll fix it before you do.

Anyhow, let me know what you think of it.

You can see the [Before](http://garno-screenshots.s3.amazonaws.com/graphql-docs-mobile-before.png) & [After](http://garno-screenshots.s3.amazonaws.com/graphql-docs-mobile-after.png) here from Chrome on desktop.

### DIFF
--- a/site/_core/Site.js
+++ b/site/_core/Site.js
@@ -21,7 +21,7 @@ export default ({ page, category, title, section, className, noSearch, children 
           `${title} | ${category || 'GraphQL'}` :
           `GraphQL | ${SiteData.description}`}
       </title>
-      <meta name="viewport" content="width=640" />
+      <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0" />
       <meta property="og:title" content="GraphQL: A query language for APIs." />
       <meta property="og:description" content="GraphQL provides a complete description of the data in your API, gives clients the power to ask for exactly what they need and nothing more, makes it easier to evolve APIs over time, and enables powerful developer tools." />
       <meta property="og:type" content="website" />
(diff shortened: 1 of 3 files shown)


leebyron:
Sorry for the delay reviewing this, this is great thanks for your help!

My high level feedback is that there are already media queries throughout the css file that would be great to leverage rather than adding new ones. In general most of the rules are written for small screens first and add additional rules via media queries for larger screens. Perhaps we could edit those existing initial rules.

Also - is it possible to keep the 640px width? After some testing I found that preserved as much of the design as possible. Though I like the idea of changing the font size for that smaller screen size just to make the actual physical font size appear larger since most devices will scale that 640 down.

I'm not seeing the horizontal scrolling issue - can you let me know what hardware/browser you're seeing that for?

garno:
Thanks for taking the time to take a look at this :)

The horizontal scrolling issue is happening on desktop only since the viewport is locked at 640px on mobile. Which results in a zoomed out page instead of a mobile-friendly one.

I wanted to make sure these were changes you would accept before taking too much time on this. I largely prefer the mobile experience in a flexible viewport. I agree that screenshots might not be the perfect medium to demonstrate this, you can access it live: https://enigmatic-eyrie-20762.herokuapp.com/

If you still prefer a fixed 640px viewport, you can close this issue.

Have a nice day!

wincent:
I'm going to close this due to inactivity (and because the problem here doesn't manifest on mobile, only on desktop in a relatively rare scenario), but feel free to comment if this is something you'd like to keep working on and we can re-open it. Thanks for the contribution, @garno!

PR #201, opened 2016-11-06
Update Community-Events.md

gsans:
Added few talks.


stubailo:
I'm not sure about this one - should we have talks that have already happened on the site? Curious what @theadactyl thinks about this one.

gsans:
I think that's how it's being use right now as you don't remove old events. The first event of the current list is from 15th September. 

I also think the ordering should be reversed if you don't remove old events.

theadactyl:
I think we should be removing old events for sure. Will take another look
at this PR and how we should be maintaining event listings later today.

leebyron:
I agree with @stubailo - let's keep this page reserved for future events! Not a lot to gain by linking to events that have already occurred.

If there are high quality videos introducing new content, then we could add those to the community section on videos?

PR #207, opened 2016-11-11
Adding a graphql subscriptions tutorial to the blog page.

mlp5ab:
We recently published this tutorial on how to build realtime apps with graphql subscriptions and it would awesome to have it included here as well. I have broken out the pull requests for adding the blog to the community page.

### DIFF
--- a/site/blog/20151110-build-realtime-apps-with-graphql-subscriptions.md
+++ b/site/blog/20151110-build-realtime-apps-with-graphql-subscriptions.md
@@ -1,0 +1,554 @@
+---
+title:  "How to Build Realtime Apps with GraphQL Subscriptions"
+layout: ../_core/BlogLayout
+permalink: /blog/how-to-build-realtime-apps-with-graphql-subscriptions/
+date: 10 Nov 2016
+byline: "Michael Paris"
+guestBio: CEO of Scaphold.io
+---
+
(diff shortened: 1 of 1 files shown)


site/blog/20151110-build-realtime-apps-with-graphql-subscriptions.md
```diff
@@ -0,0 +1,554 @@
...
+}]);
+```
+
+This auth middleware looks for a token names 'scaphold_user_token' in the browser's local storage and then appends the necessary auth header to each request. This is great because all we need to do is pick a authentication mechanism and then save the token we get back from logging in to local storage and the middleware will handle the rest.
+
+#### Social authentication for any application
+
+**Slackr** uses Auth0 lock for easy authentication. To make this work, we must first add the Auth0 integration to our API on Scaphold. It's easy!
+
+1) Go to [Auth0](https://auth0.com) and create a free account.
```
j0k3r: Can't this be converted to a native list?

leebyron:
This is awesome, but I don't think the graphql team blog is the appropriate place for publishing this kind of content. I think you should post this on your team's blog or medium!

We want to ensure that blog stays open as a channel for updates from the project and while subscriptions are an exciting future prospect, they're still not part of the graphql specification and we don't want to create confusion.

PR #216, opened 2016-11-23
Document about documentations in Schema language

dtinth:
I struggled to find ways to add a documentation for schemas generated using `buildSchema`, so that they appears GraphiQL. Without documentation it says that everything is “self-explanatory.”

I couldn’t find the docs that mentions how to do this, but [I found it in the source code](https://github.com/graphql/graphql-js/blob/36cd1622cad12ff63b01752e09e4a274b48a9d7b/src/utilities/buildASTSchema.js#L483-L487). Might be a good idea to surface it into the documentation.

I think the sentence could be made more concise and clear, but I don’t have any ideas now. Suggestions are welcome. :D

### DIFF
--- a/site/learn/Learn-Schema.md
+++ b/site/learn/Learn-Schema.md
@@ -35,19 +35,25 @@ Every GraphQL service defines a set of types which completely describe the set o
 
 GraphQL services can be written in any language. Since we can't rely on a specific programming language syntax, like JavaScript, to talk about GraphQL schemas, we'll define our own simple language. We'll use the "GraphQL schema language" - it's similar to the query language, and allows us to talk about GraphQL schemas in a language-agnostic way.
 
+In addition, some tools (including the [reference implementation](http://graphql.org/graphql-js/)) can [generate a schema from the GraphQL schema language](http://graphql.org/graphql-js/utilities/#buildschema).
+
 ### Object types and fields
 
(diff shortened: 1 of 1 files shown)


stubailo:
The specific syntax for documentation is still in flight here: https://github.com/facebook/graphql/pull/90

So I don't know if we should document that if it's not fully specified yet. Thoughts @leebyron?

leebyron:
As long as we have very clear messaging at the top that it is experimental and likely to change, I think documentation is only a good thing

leebyron:
However I'm not sure this specific change is warranted. Considering that descriptions in particular are still up for debate. Also, this page is not about the schema language, it's about the type system itself. So this is the wrong point in the course of knowledge transfer to be talking about these specific points

leebyron:
Closing this since it's not the appropriate place to discuss details of schema language, but would definitely consider docs alongside GraphQL-js

dtinth:
I see. Should I make a PR on graphql-js's API docs page instead? If so, would you prefer I reuse the PR or make a new one? I will make it clear about it being experimental.

leebyron:
Either is fine :) I can reopen if you want to repurpose this one

stubailo:
For the record I think it would be great to have docs for the schema language once it's merged into the spec, separate from GraphQL.js!

PR #224, opened 2016-12-01
Add code example for graphql-dotnet

joemcbride:


site/code/index.html.js
```diff
@@ -299,8 +299,82 @@ Code that executes a hello world GraphQL query with \`graphql-clj\`:
...
 
-  - [graphql-dotnet](https://github.com/graphql-dotnet/graphql-dotnet): GraphQL for .NET
-  - [graphql-net](https://github.com/ckimes89/graphql-net): Convert GraphQL to IQuerable
+[graphql-dotnet](https://github.com/graphql-dotnet/graphql-dotnet)
+
+A .NET library that provides a GraphQL implementation with .NET Core support.
+
+Code that executes a hello world GraphQL query with \`graphql-dotnet\`:
+
+\`\`\`csharp
```
stubailo: Why the backslashes here?

joemcbride: This is because the markdown is inlined in a JavaScript multiline string template literal so have to escape the markdown code block syntax.  See line 20.

leebyron:
This is great, but pretty long compared to some of the other getting started sections - could you trim this down to the smallest example for those getting started? Or if that's not possible, perhaps a link to this sort of example on the graphql-dotnet docs would be better?

wincent:
It's been more than a month since the last activity on this PR so I'm going to close it. Please feel free to submit a new PR, @joemcbride, once you've addressed the feedback. Thanks for the contribution!

PR #233, opened 2016-12-16
Updated Meteor URL on /users

danahmadi:
We changed the location of our blog-- just updating this link to reflect that.


wincent:
Thanks, @danahmadi!

PR #242, opened 2016-12-31
Apollo Server renamed to GraphQL Server as of v0.4

ericvangeem:
Per http://dev.apollodata.com/tools/graphql-server/migration-graphql-server.html

> In version 0.4, Apollo Server has been renamed to GraphQL Server, to reduce confusion about its compatibility with other GraphQL-related packages and tools. In particular, the goal is to ensure people don’t get confused that Apollo Client only works with Apollo Server and the other way around.

### DIFF
--- a/site/code/index.html.js
+++ b/site/code/index.html.js
@@ -87,11 +87,11 @@ app.use('/graphql', graphqlHTTP({
 app.listen(4000, () => console.log('Now browse to localhost:4000/graphql'));
 \`\`\`
 
-#### [GraphQL Server](http://dev.apollodata.com/tools/graphql-server/index.html) ([github](https://github.com/apollostack/graphql-server)) ([npm](https://www.npmjs.com/package/graphql-server))
+#### [graphql-server](http://dev.apollodata.com/tools/graphql-server/index.html) ([github](https://github.com/apollostack/graphql-server)) ([npm](https://www.npmjs.com/package/graphql-server))
 
-A GraphQL server that works with Node.js.
+A set of GraphQL server packages from Apollo that work with various Node.js HTTP frameworks (Express, Connect, Hapi, Koa etc).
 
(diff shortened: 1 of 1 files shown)


ericvangeem:
Thanks @stubailo, I've updated the PR.

wincent:
Thanks for the PR, @ericvangeem.

> Apollo Server has been renamed to GraphQL Server, to reduce confusion about its compatibility with other GraphQL-related packages and tools.

Sounds like a good move. In the context of this document, however, this might risk another kind of confusion, which is that people might think that it is somehow "the" GraphQL server.

What do you think about mentioning "Apollo" somewhere in the copy to avoid that kind of confusion? And relatedly, I think there's a missed opportunity in here to highlight that you actually have a set of packages designed to work with other server/middleware projects beyond Express.

At the moment this reads like:

> # GraphQL Server (github) (npm)
>  
>  A GraphQL server that works with Node.js.
>
> To run a hello world server with GraphQL Server:
>
> ...

How about if you made it read something like this?

> # graphql-server (github) (npm)
>  
>  A set of GraphQL server packages from Apollo that work with various Node.js HTTP frameworks (Express, Connect, Hapi, Koa etc).
>
> To run a hello world server with graphql-server-express:
>
> ...

Edits are:

- Change the title from "GraphQL Server" to "graphql-server", to match the main package name on NPM (this is what we do with express-graphql, and we should probably do the same with "GraphQL.js" at the top of this section, for consistency).
- Mention "from Apollo".
- Mention some framework variants.
- Make it explicit that the example refers to graphql-server-express.

Thoughts?

stubailo:
That's a great set of edits!

ericvangeem:
Agreed, @wincent. That sounds much better. Thanks for the input!

wincent:
Thanks, @ericvangeem. Merging.

PR #244, opened 2017-01-10
🏃 Optimize svg files

lorensr:
`svgo` removes metadata, whitespace, etc

https://github.com/svg/svgo

image


wincent:
Thanks for this @lorensr. A one-off fix like this is useful.

I wonder whether it is also worth having something integrated into the build itself so that any new images added in the future get the same treatment. Any thoughts on that?

lorensr:
never seen that, sounds nice

leebyron:
❤️

PR #247, opened 2017-01-18
Added a book to "More Resources"

packtpartner:
Added a new book link to "More Resources"


wincent:
Thanks for the PR, @packtpartner. I think we probably need a separate "Books" heading, as the section under which you've added this link leads with:

> To explore other community-developed resources and content about GraphQL, take a look at these sites:

(And your link is really about the book, not so much a site as a destination in itself.)

My suggestion would be to start a new "Books" heading, probably immediately above "More Resources", and add your link there instead.

packtpartner:
Thanks for the suggestion @wincent . I will close this PR and create a new one with a Books section.

PR #249, opened 2017-01-20
Added smarkets logo and JS object to GQL users

caglar-senel:
Also removed a single blank line that was there for some reason


wincent:
Thanks, @caglar-senel. Will merge.

PR #253, opened 2017-01-31
Add DueDil to users

webb04:


wincent:
Thanks for this, @webb04. I added a trailing slash, for consistency with the other links in the file, but otherwise this looks good.

PR #267, opened 2017-02-12
Add a bit more characters and fix appears

soswow:
I know it is a bit silly. When I was reading documentation and I tried `appearsIn` for R2-D2 my non existing OCD suddenly cried out in terror.
Sure, there are still missing droids (K-2SO) and characters missing, but at least R2-D2 will have right movie list.
Not sure if it breaks any documentation dependencies/references.

### DIFF
--- a/site/_core/swapiSchema.js
+++ b/site/_core/swapiSchema.js
@@ -50,6 +50,12 @@ enum Episode {
 
   # Star Wars Episode VI: Return of the Jedi, released in 1983.
   JEDI
+
+  # Star Wars: The Force Awakens, released in 2015.
+  FORCE
+
+  # Rogue One: A Star Wars Story, released in 2016.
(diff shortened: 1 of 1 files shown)


wincent:
It's kind of funny to even be discussing this, but the comment in the schema already indicates that it refers to the original trilogy (not even the Episode I-III are included). If we go beyond that, we'd eventually wind up with 9 films (Episodes I-IX), and even more if we include the "Star Wars Story" movies like Rogue, which can be expected to come in at a rhythm of once every two years, and likely even faster later on.

At that point, this will no longer be an educational example and instead will be a geeking-out exercise which would be fun for us Star Wars nerds but would detract from the real purpose of the material.

So, respectfully, I'm going to thank you for your contribution but I think the right course here is to close.

PR #268, opened 2017-02-13
Replace &mdash; with —

RileyTomasek:
The &mdash; is rendering correctly from the markdown. It should be an — character instead.


wincent:
Thanks, @RileyTomasek!

PR #272, opened 2017-02-18
Footer link correction

kalinchernev:
The footer link should lead to:
http://graphql.org/code/#server-libraries
Instead of:
http://graphql.org/code/#graphql-server-libraries


wincent:
Good catch! Thanks a lot @kalinchernev.

PR #273, opened 2017-02-20
Add www.hijup.com to the users

fikriauliya:
HIJUP.com is a modest fashion e-commerce.
SInce a few months ago, we switched most of our REST API to GraphQL === "Happy dev".


wincent:
Thanks, @fikriauliya.

PR #282, opened 2017-03-09
Adding GraphQL World to Users

vning93:
Adding [GraphQL World](https://graphql-world.com) to Users page.


stubailo:
We're only including each company once here - you'll note that there is only an entry for Meteor and none for Apollo.

vning93:
Sure no problem. Thanks.

PR #287, opened 2017-03-22
Add one more valid case

jferrettiboke:


stubailo:
Thanks, this is a really helpful clarification!

wincent:
@stubailo: Next time, we should wait for the CLA to be signed before merging. @jferrettiboke: care to go back and sign it retroactively?

stubailo:
@wincent my mistake! On our repos we have a failing status when CLA is not signed so it's not mergeable, might be useful to add that to the Facebook bot. I'll watch out for it next time.

PR #290, opened 2017-03-28
Update Twitter link from deprecated apollostack to current apollographql

wincent:
Just noticed this while reviewing another PR.


stubailo:
Thanks!

PR #300, opened 2017-04-24
Update graphql-clj Clojure example

tendant:


wincent:
Thanks @tendant. Not a Clojure myself, but this looks reasonable.

PR #311, opened 2017-05-10
Corrected misspelling of "dice"

cjinpheow:
Was spelt as "die".

### DIFF
--- a/site/graphql-js/Tutorial-ObjectTypes.md
+++ b/site/graphql-js/Tutorial-ObjectTypes.md
@@ -16,22 +16,22 @@ type Query {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


wincent:
Thanks for the PR @cjinpheow, but I think this is intentional: "die is the singular" form, "dice" is plural. Let me know if you think there are places where we actually meant to use the plural.

cjinpheow:
I see. Although if you perform a google on "Dice vs Die", you will see articles that say the word "Dice" is now accepted as both plural and singular. (See https://en.oxforddictionaries.com/definition/dice).  One would say, "Roll the dice", instead of "Roll the die" which sounds a bit morbid, IMHO!  But I suppose if we take the purist view, then "die' it is. Thanks!

PR #320, opened 2017-06-01
Update GitHub link to reflect move out of early access

hollenberry:
Hey there!

It looks like the GitHub URL is still pointing to the Early Access link for GraphQL. Per last week's Satellite announcement, v4 is now officially live on the documentation site. 

I'm not sure if they plan to deprecate later, but I thought I'd submit this quick fix in case they do. The v4 link should be more permanent since the v3 RESTful link remains unchanged and is the source of truth. 

Blog post of the official announcement:
screen shot 2017-05-31 at 11 21 14 pm

### DIFF
--- a/site/users/index.html.js
+++ b/site/users/index.html.js
@@ -29,7 +29,7 @@ var logos = [
     name: 'GitHub',
     img: 'github.png',
     isRound: true,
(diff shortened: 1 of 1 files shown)


wincent:
Great! Thanks @hollenberry.

PR #332, opened 2017-06-25
Add graphql-erlang to server implementations

jlouis:
Add a reference to ShopGun's GraphQL implementation for the server-side (written in Erlang)

As an aside: perhaps we should sort the implementation list based on the programming language. Right now, things are just added in the bottom it seems, so I've done that for this PR, but it feels a bit odd going through an index with no order :)


jlouis:
I submitted a CLA, but there were no "acknowledgement" of the submission, so I'm not sure it went through as expected. If not, please get back to me (or patch yourself :P)

jlouis:
Poor overworked bot :)

wincent:
The bot is a slacker.

Thanks for the PR @jlouis! Will get this merged.

wincent:
@jlouis: re: your point:

> perhaps we should sort the implementation list based on the programming language. Right now, things are just added in the bottom it seems, so I've done that for this PR, but it feels a bit odd going through an index with no order

I think this should be addressed by: https://github.com/graphql/graphql.github.io/pull/333

PR #340, opened 2017-07-03
adds GraphQLList to the imports when mentioned

bdougie:
Noticed that GraphQLList was being used but not imported in the first example. Looking forward to using this example at work to as part of some research and development I am doing at work


wincent:
Excellent. Great catch. Thanks, @bdougie!

PR #343, opened 2017-07-06
Added etmdb logo

dawitnida:


wincent:
Thanks @dawitnida. I'll merge this and then #344 (although ideally they would have been just one PR).

PR #348, opened 2017-07-09
Add githubql project, the first Go GraphQL client.

shurcooL:
To my knowledge, this is the first GraphQL client package available for [Go](https://golang.org/).

It's still in active development, but by now, it's feature complete and functional. It has support for performing queries and mutations, including arguments, variables, aliases, inline fragments, unions, etc.

The focus of the package was to come up with a friendly, simple, powerful API for a Go GraphQL client. The history of its development and API design decisions can be seen at https://github.com/google/go-github/issues/646.

Everything specific to GitHub's GraphQL API v4 is generated from the schema, so it can be easily adapted for other GraphQL APIs. It's planned to generalize the GraphQL client generator.

It has been tested via simple test programs, as well as more advanced use cases, and comes with thorough test coverage. Its API is becoming more stable, so I think it's a good time to add it to the list of GraphQL-related code.

### DIFF
--- a/site/code/index.html.js
+++ b/site/code/index.html.js
@@ -365,6 +365,7 @@ Executor.execute(schema, query) map println
 ## GraphQL Clients
 
 - [C# / .NET](#c-net-1)
+- [Go](#go-1)
 - [Java / Android](#java-android)
 - [JavaScript](#javascript-1)
 - [Swift / Objective-C iOS](#swift-objective-c-ios)
@@ -372,6 +373,10 @@ Executor.execute(schema, query) map println
 ### C# / .NET
 
   - [graphql-net-client](https://github.com/bkniffler/graphql-net-client): Basic example GraphQL client for .NET.
+
+### Go
+
+  - [githubql](https://github.com/shurcooL/githubql#readme): GraphQL client for accessing GitHub's GraphQL API, in Go, generated from schema.
 
 ### Java / Android
 


stubailo:
Hmm, this would be the first time the list would include something specific to a particular GraphQL API - adding it to a list of generic GraphQL clients seems a bit misleading.

shurcooL:
One of the goals for the package is to create a generic Go GraphQL client generator. You'd give it a name and GraphQL schema, and it would generate everything.

However, I find it easier to develop a Go GraphQL client when targeting a real GraphQL server that I can test against.

So yes, `githubql` itself is specific to a particular GraphQL API, but the generator driving it, and its API are general. It's a little too early for me to try to factor out the common/general part, but I am thinking about it.

Do you think we can resolve this issue with better phrasing? I see that some other clients are described as an "example GraphQL client". Can we say `githubql` serves as an "example GraphQL client  written in Go" or something like that?

wincent:
> Do you think we can resolve this issue with better phrasing?

No, I don't think phrasing alone can cover this. It would really need to be an entirely separate section, at which point we'd just be doing a poor job of replicating what already exists in [awesome-graphql](https://github.com/chentsulin/awesome-graphql) — see the "Examples" section there, where they already have [some Go examples](https://github.com/chentsulin/awesome-graphql#example-go). What do you think about submitting the PR there instead, @shurcooL?

In the meantime I'm going to close this one. If you do go ahead and factor out a generic GraphQL client library in Go then we'd certainly welcome a PR adding a link to that.

shurcooL:
> No, I don't think phrasing alone can cover this. It would really need to be an entirely separate section
> ...
> In the meantime I'm going to close this one. If you do go ahead and factor out a generic GraphQL client library in Go then we'd certainly welcome a PR adding a link to that.

That makes sense. Thanks for consideration. When I do factor out a generic GraphQL client library in Go, I'll send a followup PR here.

> already exists in [awesome-graphql](https://github.com/chentsulin/awesome-graphql) — see the "Examples" section there, where they already have [some Go examples](https://github.com/chentsulin/awesome-graphql#example-go). What do you think about submitting the PR there instead, @shurcooL?

They seem to only have examples for GraphQL servers in Go, nothing for clients. But I'm not sure if a GitHub-specific client will fit there either. I'll look more into it and consider it. Thanks.

stubailo:
@shurcooL I'd be happy to add it to `awesome-graphql` for sure!

PR #355, opened 2017-07-19
Reference graphql-request instead of Lokka

nikolasburk:
Updated JS section to reference [graphql-request](https://github.com/graphcool/graphql-request) instead of Lokka as work on Lokka is discontinued.

### DIFF
--- a/site/code/index.html.js
+++ b/site/code/index.html.js
@@ -391,7 +391,7 @@ Executor.execute(schema, query) map println
 
   - [Relay](https://facebook.github.io/relay/) ([github](https://github.com/facebook/relay)) ([npm](https://www.npmjs.com/package/react-relay)): Facebook's framework for building React applications that talk to a GraphQL backend.
   - [Apollo Client](http://dev.apollodata.com/react/) ([github](https://github.com/apollostack/apollo-client)): A powerful JavaScript GraphQL client, designed to work well with React, React Native, Angular 2, or just plain JavaScript.
-  - [Lokka](https://github.com/kadirahq/lokka): A simple JavaScript GraphQL client that works in all JavaScript environments - the browser, Node.js, and React Native.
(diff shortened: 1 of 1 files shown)


wincent:
Thanks for this @nikolasburk. I think I'd prefer to *add* this to the list rather than remove Lokka (which still works and is useful, despite not being committed to since November 2016; also, has there been any formal deprecation announcement from people working on Lokka?). You could move Lokka to the bottom of the list if you wanted to de-emphasize it, though. Thoughts?

stubailo:
I'd be in favor of removing Lokka personally, since it has some oddities that don't align well with modern GraphQL. For example, randomly generating fragment names.

nikolasburk:
@wincent The last thing we heard from the guys at kadira is that they moved on to other things and didn't plan to continue working on Lokka, however, I'm indeed not 100% sure about that. And I think it's reasonable to keep it in the list as long as it provides value to people (though I agree with Sashko that it's not the best choice if what you really want is just a lightweight and simple wrapper on top of `fetch`). I created an [issue](https://github.com/kadirahq/lokka/issues/41) in the project asking for an official statement on the future of Lokka and will update the PR to keep it in the list for now. Does that sound all right?

wincent:
> I created an issue in the project asking for an official statement on the future of Lokka and will update the PR to keep it in the list for now. Does that sound all right?

Sounds like the right approach to me.

> I'd be in favor of removing Lokka personally, since it has some oddities that don't align well with modern GraphQL. For example, randomly generating fragment names.

I'd prefer us to default to being more liberal in what we accept, as opposed to trying to present a very narrowly curated view of "state of the art", "best practice", "blessed/annointed" projects. Those kind of judgments may be subjective and allege a consensus that isn't real yet. We run the risk of creating an orthodoxy (not a good thing in the long term; it could stifle innovation) rather than inclusive, healthy community. Would also be possible to go too far in the liberal direction, but I don't think we're anywhere near there yet; like all things, shoot for moderation and balance...

wincent:
Just pushed a tiny commit on top of this because I want to see Travis run (failed from some reason on the last run; not yet sure if spurious or actually broken).

PR #362, opened 2017-08-03
Add link to How to GraphQL tutorial

schickling:
I love the "learn" section of graphql.org but we often hear that people want to have a more practical getting started experience. Hopefully this is the right location to mention [How to GraphQL](https://www.howtographql.com/).


leebyron:
This works for me

PR #366, opened 2017-08-05
Fix misleading title in APIReference-TypeSystem.md

hikerpig:
This plural form can kind of misleading in API section


wincent:
Good call. Thanks, @hikerpig.

Issue #283, opened 2017-03-10
Mention Scaphold.io on http://graphql.org/code/


Mitranim:
https://scaphold.io looks like a fantastic BaaS dedicated to GraphQL. I only heard about it by accident. Be nice if the official docs suggested it.

stubailo:
It doesn't fit well with the current content of that page, but it maybe could be reasonable to add a section with something like "GraphQL services"? I think there are like 5-6 services that would qualify.

leebyron:
Would consider PRs adding this kind of section to that page!

Bonus points for a nice nav sidebar that helps people jump around on that page

christiane-deneser:
Hi there,
can I work on that issue and add a nav sidebar to the page "Code"?

Would you consider dividing the page in three parts?: 
1. Servers 
2. Clients 
3. Tools & Services (including "More Stuff")?
In that way the wording in the footer nav is (more or less) matched.

Or would you prefer a fourth part for Services (including "More Stuff")?
In that way I would add another link "Services" to the footer nav as well.

Also, I could add a "continue reading" link at the end of each part. That way the page becomes more structured and easier to perceive as lots of content has accumulated over the past months.

I would be happy to help. Also, I am pretty new to contributing to OS and find that issue a nice way to start.

Best,
Christiane

Mitranim:
Nevermind, the service went down.

PR #368, opened 2017-08-07
List prominent GraphQL service providers on the code page

Sonicrida:
I was attempting to assist with the request mentioned in #283 . This is my first time contributing to an open source project so feel free to let me know if I've done something incorrectly 😄. @stubailo mentioned 5-6 different services but I could only find these 4 just by searching for GraphQL services. I wasn't too sure about how to order them so I figured that I would list Graphcool and Scaphold first since those are more targeted towards general usage while the other two are more specific in their use cases.


stubailo:
Would also appreciate a mention of http://www.apollodata.com/optics/, a GraphQL-specific performance tool.

Also, Reindex doesn't offer the GraphQL service anymore: http://reindex.io/

Sonicrida:
Good call on Apollo Optics! Also, I believe that Reindex is still offering their service at this url if I'm not mistaken: https://www.reindex.io/baas/ It seems that I am still able to sign up.

wincent:
Thanks a bunch @Sonicrida. Will merge with a few small tweaks applied on top.

Sonicrida:
Thanks @wincent .That's awesome! I'm glad to help!

PR #376, opened 2017-08-17
Add ComparaOnline to users list

joaquinrulin:


wincent:
Thanks @joaquinrulin.

PR #377, opened 2017-08-21
Add Goalify to users list

trongthanh:


wincent:
Thank you @trongthanh.

PR #381, opened 2017-09-02
Happy new year

greenstatic:
Automatically changes the year in the footer of the page, with fallback to the year 2017 if JS is disabled.


leebyron:
Works for me

PR #383, opened 2017-09-06
Patch 1

xgqfrms-gildata:


xgqfrms-gildata:
OK

wincent:
Thanks for the PR but I'm afraid that I don't feel comfortable merging this. It's not even clear what the linked website is for, and I can't tell from looking at [the corresponding GitHub repo](https://github.com/xgqfrms/xgqfrms) or [owner profile](https://github.com/xgqfrms) either. The PR title ("Patch 1") and missing description doesn't shed any light either. The supplied graphic is riddled with compression artifacts etc.

PR #387, opened 2017-09-15
Add reference to GQL

sdelamo:


wincent:
Thanks @sdelamo. Will fix the heading level and merge.

PR #398, opened 2017-10-03
Change `GraphQLInputObjectTypeConfig` to `GraphQLInputObjectConfig`

learnercys:


wincent:
Thanks @learnercys. Are you sure about this change? I think graphql defines only `GraphQLInputObjectTypeConfig`, not `GraphQLInputObjectTypeConfig`.

learnercys:
@wincent I think I'm because there's no reference for `GraphQLInputObjectTypeConfig` and the `GraphQLInputObjectType` section follows the same pattern to describe its configuration so I asume that the right reference for the InputObjectType Configuration is `GraphQLInputObjectConfig` instead. I'll let a description for that.

PR #405, opened 2017-10-27
Add mongo-graphql-schema to the Code page

arackaf:
Hi.  This PR adds my mongo-graphql-schema to the Code page.  Hopefully it's a good fit :)

### DIFF
--- a/site/code/index.html.js
+++ b/site/code/index.html.js
@@ -33,6 +33,7 @@ In addition to the GraphQL [reference implementations in JavaScript](#javascript
 - [Groovy](#groovy)
 - [Java](#java)
 - [JavaScript](#javascript)
+- [MongoDB](#mongodb)
 - [PHP](#php)
 - [Python](#python)
 - [Scala](#scala)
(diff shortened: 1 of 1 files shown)


arackaf:
Hi - the CLA has been signed - I don't see any update comments, but it should be signed now.  Let me know if it's somehow not and I'd be happy to do whatever else I need :)

stubailo:
@arackaf thanks for the PR, but this doesn't seem like a good fit for this list, which is more low-level implementations rather than starter kits or higher level libraries. I think this library would fit better on https://github.com/chentsulin/awesome-graphql

arackaf:
@stubailo thanks - but are you sure?  Mine seems like the Mongo equivalent of this (which is also listed on the same page)

https://github.com/graphql-dotnet/graphql-dotnet 

if not no biggie :)

stubailo:
I haven't looked in detail, but I suspect that project is the core implementation of GraphQL for C#, which isn't specific to a particular database, and doesn't use a separate implementation. .NET is more like Node.js than it is like MongoDB.

Issue #241, opened 2016-12-29
operationName is hard to grasp


martinheidegger:
The current section about operation names, here: http://graphql.org/learn/queries/#operation-name
reads like this: 

> One thing we also saw in the example above is that our query has acquired an operation name. Up until now, we have been using a shorthand syntax where we omit both the query keyword and the query name, but in production apps it's useful to use these to make our code less ambiguous.
>
> Think of this just like a function name in your favorite programming language. For example, in JavaScript we can easily work only with anonymous functions, but when we give a function a name, it's easier to track it down, debug our code, and log when it's called. In the same way, GraphQL query and mutation names, along with fragment names, can be a useful debugging tool on the server side to identify different GraphQL requests.

The problem is that after reading this section there is no example or other hint on the shape in which the operation's name should be specified or where to look up the spec. It would be nice to be pointed to somewhere in this section.

Aside from that: Thank you for your great work!

wincent:
> The problem is that after reading this section there is no example or other hint on the shape in which the operation's name should be specified or where to look up the spec. It would be nice to be pointed to somewhere in this section.

Good point, @martinheidegger. Would you like to submit a PR that adds a (probably small) example?

stubailo:
I agree with this - that section introduces a lot of new concepts at the same time and should probably do that incrementally.

martinheidegger:
@wincent Sorry, but I havn't figured out how it is supposed to work yet.

stubailo:
Operation name is a "cosmetic" name for this particular query, much like a function name in JavaScript. So:

```graphql
{ me { username }}
```

Is like an anonymous function:

```js
() => { return 'hello' }
```

Here's a query with an operation name:

```graphql
query MeQuery {
  me { username }
}
```

That's like a named function:

```js
function getGreeting() {
  return 'hello';
}
```

Basically it's a string you can use to refer to that function in your code, developer tools, logs, stack traces etc.

idibidiart:
@stubailo a funny thing I ran into just now that's confusing m noob brain:

when I don't use "query" as in "query { ... " and instead use just "{ ... " GraphiQL returns an error saying it "Unknown operation named \"null\"" 

If I use "query { ..." it works just fine, but if I then remove "query" and re-evaluate it works, too, but not after I refresh the GraphiQL page (then it requires 'query' qualifier again or it produces the aforementioned error.

I guess a bug in GraphiQL?

wincent:
[...]

PR #407, opened 2017-10-30
Section operation name: provide code example, edit text, change order…

christiane-deneser:
… of topics

Hi there,
in my PR, I worked on Issue #241.
* I added a code example for operation names
* and edited the texts, both referring to the suggested article ( https://dev-blog.apollodata.com/the-anatomy-of-a-graphql-query-6dffa9e9e747).
* I changed the order of the topics: Section 'Operation name' now comes before section 'Variables'. I find that easier to grasp, as the examples in 'Variables' use an operation type and name, but neither had been explained before. Therefore I also adjusted the order of topics in the nav sidebar.

### DIFF
--- a/site/learn/Learn-Queries.md
+++ b/site/learn/Learn-Queries.md
@@ -4,7 +4,7 @@ layout: ../_core/DocsLayout
 category: Learn
 permalink: /learn/queries/
 next: /learn/schema/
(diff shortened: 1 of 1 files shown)


stubailo:
@christiane-deneser this is wonderful, thanks!

PR #421, opened 2017-11-17
Change die to dice

sebastianlp:


sebastianlp:
Ok, I saw this PR: #311 and I think we should close this.

PR #424, opened 2017-11-22
Add new meetup in Dallas / Fort Worth

pengwynn:
We're starting a new meetup in DFW next month and would like to add it to the community events listing.


stubailo:
Thank you!

PR #425, opened 2017-11-25
Clarify the distinction between rootValue and object instance

tychota:
As per https://github.com/graphql/graphql-js/issues/1098#issuecomment-346967872, it wasn't clear for me the resolver could return extra value and that nested fields could use this to help resolving them.

What would you think of clarifying this ?

Can you give me feedbacks to the proposal ?


site/learn/Learn-Execution.md
```diff
@@ -149,6 +153,52 @@ The resolver for this field is not just returning a Promise, it's returning a *l
...
+
+```js
+Starship: {
+  timeSpendInside(obj, args, context) {
+    return context.db.loadTimeSpendInsideByStarthipIdAndPiloteId(obj.id, obj._human)
+  }
+}
+```
+
+Even if doing this is a possibility, tryy to avoid relying to much on it as it increase the coupling of your API.
```
IvanGoncharov: `tryy` => `try`

tychota:
@leebyron @stubailo, can you review this PR ?

site/learn/Learn-Execution.md
```diff
@@ -149,6 +153,52 @@ The resolver for this field is not just returning a Promise, it's returning a *l
...
+Query: {
+  human(obj, args, context) {
+    return context.db.loadHumanByID(args.id).then(
+      userData => new Human(userData)
+    )
+  }
+}
+```
+
+But nothing prevent you for returning extra elements.
```
Neaox: `prevent` => `prevents`

site/learn/Learn-Execution.md
```diff
@@ -149,6 +153,52 @@ The resolver for this field is not just returning a Promise, it's returning a *l
 GraphQL will wait for all of these Promises concurrently before continuing, and when left with a list of objects, it will concurrently continue yet again to load the `name` field on each of these items.
 
 
+## Providing extra data in a resolver
+
+The resolver returns next reducer's root value. Traditionnaly, this root value would match the instance of the Type resolved.
```
cathy-f: Traditionnaly  => Traditionally

site/learn/Learn-Execution.md
```diff
@@ -70,7 +74,7 @@ Query: {
 
 This example is written in JavaScript, however GraphQL servers can be built in [many different languages](/code/). A resolver function receives three arguments:
 
-- `obj` The previous object, which for a field on the root Query type is often not used.
+- `obj` The previous object (also called `RootValue`), which for a field on the root Query type is often not used.
```
leebyron: it is only the root value in the case where this resolver is on the root type. e.g. a root value is provided to correspond to the Query type, but a value for Human type in this page's example is not accurately called a RootValue

site/learn/Learn-Execution.md
```diff
@@ -47,7 +49,7 @@ In order to describe what happens when a query is executed, let's use an example
 }
 ```
 
-You can think of each field in a GraphQL query as a function or method of the previous type which returns the next type. In fact, this is exactly how GraphQL works. Each field on each type is backed by a function called the *resolver* which is provided by the GraphQL server developer. When a field is executed, the corresponding *resolver* is called to produce the next value.
+You can think of each field in a GraphQL query as a function or method of the previous type which returns the next type (or a subset of the next type, see **Providing extra data in a resolver** section). In fact, this is exactly how GraphQL works. Each field on each type is backed by a function called the *resolver* which is provided by the GraphQL server developer. When a field is executed, the corresponding *resolver* is called to produce the next value.
```
leebyron: I think this addition is confusing, what does it mean to return something that is a subset of the expected return type? We always expect a field that claims to return a User to return a User and not something smaller than that

site/learn/Learn-Execution.md
```diff
@@ -56,7 +58,9 @@ If a field produces a scalar value like a string or number, then the execution c
 
 At the top level of every GraphQL server is a type that represents all of the possible entry points into the GraphQL API, it's often called the *Root* type or the *Query* type.
 
-In this example, our Query type provides a field called `human` which accepts the argument `id`. The resolver function for this field likely accesses a database and then constructs and returns a `Human` object.
+In this example, our Query type provides a field called `human` which accepts the argument `id`. The resolver function for this field likely accesses a database and then constructs and returns a `Human` object. 
+
+Notice that you can also return extra properties that are not part of the instance `Human`, so deeper resolvers in your query resolution tree could take these extra data and use it to help resolve nested information (see **Providing extra data in a resolver** section).
```
leebyron: This is incorrect - a resolver cannot return extra information other than the resolved value

site/learn/Learn-Execution.md
```diff
@@ -87,7 +91,7 @@ human(obj, args, context) {
...
+The `context` is used to provide access to a database which is used to load the data for a user by the `id` provided as an argument in the GraphQL query. Since loading from a database is an asynchronous operation, this returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). In JavaScript, Promises are used to work with asynchronous values, but the same concept exists in many languages, often called *Futures*, *Tasks* or *Deferred*. When the database returns, we can construct and return a new `Human` object.  Notice that you can also return extra properties that are not part of the instance `Human`, so deeper resolvers in your query resolution tree could take these extra data and use it to help resolve nested information (see **Providing extra data in a resolver** section).
```
leebyron: I find this addition confusing

site/learn/Learn-Execution.md
```diff
@@ -149,6 +153,52 @@ The resolver for this field is not just returning a Promise, it's returning a *l
...
+  human(obj, args, context) {
+    return context.db.loadHumanByID(args.id).then(
+      userData => new Human(userData)
+    )
+  }
+}
+```
+
+But nothing prevents you for returning extra elements.
+For instance, returning in a reducer a pointer to its parent may be usefull.
```
leebyron: This is an anti pattern that we should not encourage in the docs

stubailo:
Any updates on this @tychota ?

site/learn/Learn-Execution.md
felipap: Hey @leebyron. What is the alternative to modifying the parent value? I'm having a really hard time finding any workaround for this online.

orta:
Alright, I'm going to close out this PR - it's been sitting for a year 👍 

I'm happy to merge a PR on the ID not being in the types though. Other changes will need to go through Lee

site/learn/Learn-Execution.md
kevin-krug: @felipap It seems that the anti pattern here is creating context dependent child resolvers.

PR #439, opened 2017-12-19
Fix API Reference on return type of `execute`

brabeji:
API Reference incorrectly states that return type of `execute` is always a `Promise`
https://github.com/graphql/graphql-js/blob/8173c24dba61c5e032e4abd518729a5591b3076c/src/execution/execute.js#L147


brabeji:
https://github.com/graphql/graphql-js/blob/8173c24dba61c5e032e4abd518729a5591b3076c/src/execution/execute.js#L147

PR #441, opened 2017-12-23
Clarify operation type and name requirements

jamesmacaulay:
The current text can make it seem that an operation name is required for any mutations or subscriptions, and for any operation with variable definitions:

> Up until now, we have been using a shorthand syntax where we omit both the query keyword and the query name, but in production apps it's useful to use these to make our code less ambiguous. You'll need these optional parts to a GraphQL operation if you want to execute something other than a query or pass dynamic variables.

As far as I can tell, the spec only requires operation names in multi-operation documents:

http://facebook.github.io/graphql/October2016/#sec-Language.Operations
http://facebook.github.io/graphql/October2016/#sec-Validation.Operations

This pull request modifies the text to be more precise about the requirements for operation type and name.

### DIFF
--- a/site/learn/Learn-Queries.md
+++ b/site/learn/Learn-Queries.md
@@ -126,8 +126,7 @@ You can see how the above query would be pretty repetitive if the fields were re
 
 ## Operation name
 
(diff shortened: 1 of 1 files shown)


leebyron:
thanks!

PR #456, opened 2018-01-18
More ES6 in examples

caub:


site/graphql-js/Tutorial-ObjectTypes.md
```diff
@@ -41,11 +41,7 @@ class RandomDie {
   }
 
   roll({numRolls}) {
-    var output = [];
-    for (var i = 0; i < numRolls; i++) {
-      output.push(this.rollOnce());
-    }
-    return output;
+    return Array.from({ length: numRolls }, () => this.rollOnce());
```
leebyron: Let's avoid 1Array.from({ length })` since it's a bit of an antipattern and is not easier to understand than the prior code

PR #457, opened 2018-01-22
Update Community-Events.md

komlew:
Moved two events from 2017 to “Past events” and added two upcoming 2018 events.


leebyron:
Sorry for not seeing your PR before.. this appears to be far out of date so I'm closing it now, but I'll try to do better in the future to get to changes like these in a more timely manner

PR #458, opened 2018-01-24
Cleanup upcoming events

nikolasburk:


leebyron:
Lol yep - we should probably ditch this page since it rots quickly

nikolasburk:
I don't mind updating it every once in a while but that's probably indeed not the most scalable approach to maintain it...  🙃  Can you take a look at the most recent PR for it as well @leebyron https://github.com/graphql/graphql.github.io/pull/483

PR #463, opened 2018-01-31
Added links and description for Elide implementation of GraphQL

aklish:


aklish:
@leebyron Thanks for the feedback.  I added a small code sample and modified the description slightly.  Let me know if you want more changes.

PR #528, opened 2018-06-15
Change Ruby example to print JSON instead of object

neudabei:
I wondered whether changing the Ruby example of the server side code might be useful. Converting the result of `Schema.execute` to `json` before printing it renders a more readable result: `{"data":{"hello":"Hello world!"}}` rather than just the object `#<GraphQL::Query::Result:0x00007f893535cf80>`.

### DIFF
--- a/site/code/index.html.js
+++ b/site/code/index.html.js
@@ -361,7 +361,8 @@ Schema = GraphQL::Schema.define do
   query QueryType
 end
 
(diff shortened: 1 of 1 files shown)


leebyron:
Sure, these are intended to be minimal examples, but happy to include this convenience since it's a small change

neudabei:
Makes sense. Thanks @leebyron!

PR #531, opened 2018-06-22
Object type changes

satyam22:
I believe `query` and `mutation` are name of objects. Object type should be `Query` and `Mutation`
Please let me know if i'm not correct in this matter.

### DIFF
--- a/site/learn/Learn-Schema.md
+++ b/site/learn/Learn-Schema.md
@@ -83,7 +83,7 @@ schema {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


leebyron:
This is referring to the root types as described by the schema definition, where lowercase is correct. Uppercase would usually referring to the name of the actual type provided as a query or mutation type, which is often called `Query` and `Mutation` (as it is in that page's example) but isn't required to be named that.

PR #541, opened 2018-06-29
link to variables section in fragments section

a-b-r-o-w-n:
As someone new to graphql, it was not clear to me that I could use variables inside of a fragment given the fragment example. 

This will _hopefully_ help others with similar confusion.

### DIFF
--- a/site/learn/Learn-Queries.md
+++ b/site/learn/Learn-Queries.md
@@ -124,7 +124,7 @@ fragment comparisonFields on Character {
 
 You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine lots of UI components with different fragments into one initial data fetch.
 
-### Fragments with arguments
+### Using variables inside fragments
 
 It is possible for fragments to access variables declared in the query or mutation. See [variables](learn/queries/#variables).
 
--- a/site/learn/Learn-Queries.md
+++ b/site/learn/Learn-Queries.md
@@ -122,7 +122,35 @@ fragment comparisonFields on Character {
 }
 ```
 
-You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine lots of UI components with different fragments into one initial data fetch. Much like other `type` declarations, `fragment`s can use [variables](learn/queries/#variables).
+You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine lots of UI components with different fragments into one initial data fetch.
+
+### Fragments with arguments
+
+It is possible for fragments to access variables declared in the query or mutation. See [variables](learn/queries/#variables).
+
+```graphql
+# { "graphiql": true }
+query HeroComparison($first: Int = 3) {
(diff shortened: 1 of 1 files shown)


site/learn/Learn-Queries.md
```diff
@@ -122,7 +122,7 @@ fragment comparisonFields on Character {
 }
 ```
 
-You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine lots of UI components with different fragments into one initial data fetch.
+You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine lots of UI components with different fragments into one initial data fetch. Much like other `type` declarations, `fragment`s can use [variables](learn/queries/#variables).
```
stubailo: Two comments:

1. I don't think "type declaration" is the right set of words here, since that sounds like it refers to the schema language (`type Person { ... }`) and not queries/mutations.
2. This skips over the fact that fragment variables work pretty differently than queries/mutations, since they depend on the query inside which the fragment is used. I think it might be good to add an example here.

astephenb: Thanks for the response. Is there a query example that I could borrow from with the star wars schema? I am unfamiliar with the graphql markdown.

What I would like to do is something like:

```graphql
fragment comparisonFields on Character {
  name
  appearsIn
  friends(droid: $droid) {
    name
  }
}

query HeroComparison($droid: Boolean) {
  leftComparison: hero(episode: EMPIRE) {
    ...comparisonFields
  }
  rightComparison: hero(episode: JEDI) {
    ...comparisonFields
  }
}
```

stubailo: Yeah if you look a little higher in the file you'll see how to do it:

    ```graphql
    # { "graphiql": true }
    {
      leftComparison: hero(episode: EMPIRE) {
        ...comparisonFields
      }
      rightComparison: hero(episode: JEDI) {
        ...comparisonFields
      }
    }
    
    fragment comparisonFields on Character {
      name
      appearsIn
      friends {
        name
      }
    }
    ```

astephenb: What I'm asking specifically is if I can just add a variable like `droid` in the query, or if the graphiql markdown is really driven by a schema, which would result in an error.

stubailo: ohhh, yes there is a schema, it's here: https://github.com/graphql/graphql.github.io/blob/source/site/_core/swapiSchema.js

astephenb: Thanks. I'll take a look and update my commit.

site/learn/Learn-Queries.md
```diff
@@ -124,6 +124,34 @@ fragment comparisonFields on Character {
 
 You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine lots of UI components with different fragments into one initial data fetch.
 
+### Fragments with arguments
```
stubailo: Could we call this "Using variables inside fragments"? There's an important distinction between arguments and variables.

a-b-r-o-w-n: That's a good point. I'll update.

stubailo:
Thanks!

PR #561, opened 2018-08-14
update hello world example of apollo-server-express

cooperhsiung:
The hello world example of apollo-server-express seems to be outdated, I got the lastest example from [https://www.npmjs.com/package/apollo-server-express](https://www.npmjs.com/package/apollo-server-express)


orta:
Thanks @cooperhsiung  - I've merged a PR which does this already 👍

PR #567, opened 2018-09-01
Refer Back to Self

dhurlburtusa:
I believe there is a typo. I think `Person` should be `PersonType`.


orta:
Yeah, I think you're right here - thanks

PR #574, opened 2018-09-12
Update BestPractice-Pagination.md

acatejr:
Proposed grammar fix.


orta:
Thanks for the PR @acatejr - looks like [the internet generally agrees with you](https://ell.stackexchange.com/questions/429/most-simple-or-simplest#447)

PR #579, opened 2018-09-20
Update Community-Resources.md

modysseus:
this @graphqlnews twitter account is inactive, it hasn't tweeted anything since January 2017


stubailo:
True!

PR #580, opened 2018-09-27
Update Code tab - Apollo-Server version bump is breaking the example,

ajmalmsali:
I'm not sure if we should add example from apollo-server2 instead of the version bump, anyways the example is broken as of now,


ajmalmsali:
@fbwut

orta:
I've merged a PR fixing this

PR #595, opened 2018-10-23
Update "Upcoming Events".

abernix:
I haven't removed GraphQL Europe since @dcorb has already handled that via #592, but I've added _GraphQL Conf_ and removed _GraphQL Finland_.


dcorb:
👍

PR #604, opened 2018-11-01
Be clearer about nullable items in an array

reconbot:
This bit of example code has allowed nulls in to many array types in many examples when that wasn't the intention.


stubailo:
Thanks!

PR #606, opened 2018-11-04
Add GraphQL Bangkok to Community Events

lnmunhoz:
Hello from Thailand!

We are organizing the GraphQL Bangkok since July. Would be great if we can be listed here :) 
Our next meetup is on 15th November and is our 3rd meetup already.

cc @MeixnerTobias
Thank you!


lnmunhoz:
Close duplicated #546

PR #609, opened 2018-11-07
Add Clubhouse as a user

eliperkins:
Does what it says on the tin.


eliperkins:
Closing as I no longer work for Clubhouse.

PR #613, opened 2018-11-14
Add the Alliance of American Football to users

ccbrown:
This adds the Alliance of American Football to users.


ccbrown:
Closing as the AAF has shut down and the users page doesn't seem to be maintained anyway.

PR #614, opened 2018-11-15
Rename coletica.png to correct name colectica.png

grrttedwards:
Rename the file so it doesn't 404 on the `/users` page


grrttedwards:
Dupe of #521

PR #616, opened 2018-11-16
Fix confusing indentation in /code#Clojure

lverns:
With lacinia being a bullet point and the other projects under Clojure being
headers, it looked like lacinia was nested under graphql-clj.


stubailo:
Thanks!

PR #623, opened 2018-11-28
Use new class-based API of graphql-ruby

coderliu:
[Intro of new class-based API](http://graphql-ruby.org/schema/class_based_api)


orta:
Thanks!

PR #626, opened 2018-12-11
Add Gunosy

azihsoyn:
We are using GraphQL via AWS AppSync.
https://tech.gunosy.io/entry/gunosy-sports2


caniszczyk:

graphql/graphql-landscape#48

PR #635, opened 2019-01-01
Update Learn-Schema.md -- Improve Union Type Examples

knpwrs:
This PR adds `__typename` to the query to show that union types can still be differentiated on the client and adds a second example showing common fields being queried in one spot.

There is another way to accomplish the same thing that I pointed out in this PR using fragments. Ultimately I felt like the solution targeting the common interface was cleaner and adding a note about Fragments might feel cluttered, but we can add this example if desired as well:

```graphql
{
  search(text: "an") {
    __typename
    ... on Human {
      ...CharacterProps
      height
    }
    ... on Droid {
      ...CharacterProps
      primaryFunction
    }
    ... on Starship {
      name
      length
    }
  }
}

fragment CharacterProps on Character {
  name
}
```

### DIFF
--- a/site/learn/Learn-Schema.md
+++ b/site/learn/Learn-Schema.md
@@ -311,6 +311,7 @@ In this case, if you query a field that returns the `SearchResult` union type, y
 # { "graphiql": true}
 {
   search(text: "an") {
+    __typename
     ... on Human {
       name
       height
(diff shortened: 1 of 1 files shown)


site/learn/Learn-Schema.md
```diff
@@ -311,6 +311,7 @@ In this case, if you query a field that returns the `SearchResult` union type, y
 # { "graphiql": true}
 {
   search(text: "an") {
+    __typename
```
stubailo: Can we add something to the text regarding this?

knpwrs: Sure thing!

knpwrs: @stubailo That should be all set now.

stubailo:
Thanks!

PR #636, opened 2019-01-04
Changed GraphQL.js Tutorial to arrow functions

dcavalcante:
To keep it consistent with the examples on the begging of this guide I rewrote the rest as arrow functions


orta:
OK, cool, seems reasonable to me 👍 - thanks

PR #643, opened 2019-01-18
Fix prototype of astFromValue

livingmine:
There is a mismatch between what is stated in the docs and the typescript definition. From typescript definition:
```
export function astFromValue(value: any, type: GraphQLInputType): Maybe<ValueNode>;
```


carolstran (approved):
Good eye, thanks for taking care @livingmine 🎉 

Pinging @IvanGoncharov to review/merge.

PR #646, opened 2019-01-20
Add Agoo to the list of Ruby Servers

ohler55:
Agoo with GraphQL support was just released. Agoo is a high performance server written as a C extension gem for Ruby. The GraphQL support uses straight SDL for type definitions. Ruby classes are just Ruby. They do not need to inherit from any base class. The intent is to make it easy and simple.


ohler55:
How can this be moved forward? I'd like to contribute more but that only makes sense if contributions are accepted. Please let me know if there is anything else that needs to be done to get this merged.

ohler55:
Thank you!

PR #648, opened 2019-01-22
Update Tutorial-Authentication.md

joelcolucci:
## Background 
The current GraphQL Authentication and Express Middleware documentation states the "request" object is passed as the second argument in a resolver.

However, the express-graphql (v0.7.1 ) documentation states the "request" object is passed as the third argument to resolvers. See here: https://github.com/graphql/express-graphql#combining-with-other-express-middleware

I've tested this and it appears the express-graphql documentation is correct.

## Proposed File Change
I've updated this page to match the express-graphql documentation.

Please let me know if you have any questions.

### DIFF
--- a/site/graphql-js/Tutorial-Authentication.md
+++ b/site/graphql-js/Tutorial-Authentication.md
@@ -9,7 +9,7 @@ next: /graphql-js/constructing-types/
 
 It's simple to use any Express middleware in conjunction with `express-graphql`. In particular, this is a great pattern for handling authentication.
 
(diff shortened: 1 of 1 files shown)


carolstran (approved):
Makes sense to me 🤷🏼‍♀️ Thanks for taking care @joelcolucci 🎉 

Pinging @IvanGoncharov to review/merge.

IvanGoncharov:
Resolvers inside rootValue have special behavior where all the arguments are shifted by one.
 
> However, the express-graphql (v0.7.1 ) documentation states the "request" object is passed as the third argument to resolvers.

Docs are fixed and now it states it's passed as `context` argument.

PR #665, opened 2019-02-24
Mention that there's a formal spec for the connections pattern

orta:
Step one of https://github.com/facebook/relay/pull/2603 - making sure that there's a direct route to the relay spec from here.


orta:
This is good to go

PR #666, opened 2019-02-24
Adds the object identification docs into the GraphQL best practices section

orta:
Part two of https://github.com/facebook/relay/pull/2603 - this one takes the revised spec (from that PR) for the object identification pattern and adds it to the best practices section. 

While I agree with @josephsavona fm that PR, I think making a new section just for this probably isn't worth it from an organizational perspective (within the context of this site)

Preview: https://build-malkmkyuhz.now.sh/learn/global-object-identification/


orta:
OK, after re-reading the chat in the relay issue https://github.com/facebook/relay/pull/2603#issuecomment-458218318 - I think this is good to go

Issue #692, opened 2019-04-28
Flutter GraphQL is not mentioned at website


nitishk72:
Flutter GraphQL is not mentioned at the GraphQL website.

praveenweb:
This seems to be merged via https://github.com/graphql/graphql.github.io/pull/694

@orta - We can close this?

PR #693, opened 2019-04-28
Added Flutter Graphql #692

nitishk72:
#692 Added Flutter GraohQL


nitishk72:
In this PL I added Flutter to the Server Libraries List.
I had made more PL #694 and correct this mistake.

PR #707, opened 2019-05-27
Fix grammar on BestPractice-Introduction.md

Ivaylo-Kirov:
Fixed a small typo.

### DIFF
--- a/site/learn/BestPractice-Introduction.md
+++ b/site/learn/BestPractice-Introduction.md
@@ -45,7 +45,7 @@ In contrast, GraphQL only returns the data that's explicitly requested, so new c
 
 ### Nullability
 
-Most type systems which recognise "null" provide both the common type, and the *nullable* version of that type, whereby default types do not include "null" unless explicitly declared. However, in a GraphQL type system, every field is *nullable* by default. This is because there are many things which can go awry in a networked service backed by databases and other services. A database could go down, an asynchronous action could fail, an exception could be thrown. Beyond simply system failures, authorization can often be granular, where individual fields within a request can have different authorization rules.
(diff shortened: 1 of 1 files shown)


site/learn/BestPractice-Introduction.md
```diff
@@ -45,7 +45,7 @@ In contrast, GraphQL only returns the data that's explicitly requested, so new c
...
+Most type systems which recognise "null" provide both the common type, and the *nullable* version of that type, whereby default types do not include "null" unless explicitly declared. However, in a GraphQL type system, every field is *nullable* by default. This is because there are many things which can go awry in a networked service backed by databases and other services. A database could go down, an asynchronous action could fail, an exception could be thrown. Beyond simply system failures, authorization can often be granular, where individual fields within a request can have different authorization rules.
```
carolstran, line 48: Great catch! Since we're updating this paragraph anyway, I'd also suggest...

```suggestion
Most type systems which recognise "null" provide both the common type and the *nullable* version of that type, whereby default types do not include "null" unless explicitly declared. However, in a GraphQL type system, every field is *nullable* by default. This is because there are many things that can go awry in a networked service backed by databases and other services. A database could go down, an asynchronous action could fail, an exception could be thrown. Beyond simply system failures, authorization can often be granular, where individual fields within a request can have different authorization rules.
```

Since it's hard to see the changes, here's what I did:
1. Removed the comma after `common type` in the first sentence
1. Swapped `which can go awry` for `that can go awry` for better grammar.

carolstran (approved):
Had a suggestion for other tweaks to make, but overall looks good! Pinging @IvanGoncharov to merge.

PR #714, opened 2019-06-08
temporary redirect until we figure out the swapi-graphql issue

acao:


acao:
actually, we decided on another approach, revising this, hold up!

acao:
IMHO, the redirect component should use `window.location.replace()` so as to rewrite browser history, but thats for another PR

PR #715, opened 2019-06-10
remove temporary redirect for swapi-graphql, use netlify 200 rewrite

acao:
@caniszczyk this should be a more permanent solution? what do you think?

### DIFF
--- a/site/_redirects
+++ b/site/_redirects
@@ -1,0 +1,1 @@
+/swapi-graphql  https://graphql.github.io/swapi-graphql  200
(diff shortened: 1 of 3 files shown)


site/_redirects
```diff
@@ -0,0 +1 @@
+/swapi-graphql  https://graphql.github.io/swapi-graphql  200
```
JoelMarcey: So this will do a https://graphql.github.io/swapi-graphql => https://graphql.org/swapi-graphql

acao: yes, as a 200 [netlify rewrite](https://www.netlify.com/docs/redirects/#rewrites-and-proxying), looks similar to an apache .htaccess rewrite rule or nginx `proxy_pass` rule

acao:
thanks @JoelMarcey !

PR #716, opened 2019-06-10
add trailing slashes to prevent github.io 301, splat for query params

acao:
Ope! After talking to Netlify support who's been super helpful, he pointed out that we are pointing to a githubpages path that forces a 301 redirect. Adding the trailing slash should fix this, as github.io will send a 200 for graphql.github.io/swapi-graphql/ and a 301 for graphql.github.io/swapi-graphql to the latter.

(added) a splat for QPs as well, such as `graphql.org/swapi-graphql/?query=234234234`.
this should hopefully fix legacy links, where folks have GraphQL query examples/etc that point to this domain


acao:
@JoelMarcey  @caniszczyk  this **should** be the permanent fix, if you have time to review please!

PR #717, opened 2019-06-11
better redirects using 301 HTTP code

acao:
handy as a point of reference, and cleans up the filesystem a bit too

now that we have `_redirects` in place and we know its working, I thought I'd go all out.

there was _almost_ a chance to do this more dynamically, but there were enough subtle differences that i made it very static

believe it or not, I read somewhere that google does indeed handle these soft JS redirects like they're 301's sometimes! poor google :/

should I also drop the Redirect.js react component, because it's no longer used? maybe document somewhere in Contributing.md how to add redirects too?


acao:
@orta fyi this is a pretty straightforward PR that should reduce the bloat of these additional files/directories

orta:
Perfect, let's get this in.

PR #739, opened 2019-07-18
appointy/jaal - develop spec compliant GraphQL servers in go

anujdecoder:
Added repository [Jaal](https://github.com/appointy/jaal) in Go server libraries list.


orta:
Thanks

PR #741, opened 2019-07-22
Junte company

slapshin:
We have recently started using GraphQL but this technology is very inspiring to us, it has solved a lot of problems encountered in the implementation of client-server interaction: several types of DTO, a lot of parallel queries, complex logic to add additional fields to the DTO... In the near future, we plan to fully switch to GraphQL in current projects


caniszczyk:

graphql/graphql-landscape#48

PR #745, opened 2019-07-26
Added three companies + improved image sizes

swernerx:
This adds three companies which use GraphQL here in Germany.

I also saw that some of the logos were far too large and some still had very large file sizes. So I reduced some images to <=1000px and had ImageOptim optimize them all.


caniszczyk:

graphql/graphql-landscape#48

PR #750, opened 2019-08-08
Added IBM as a GraphQL User.

laredo:
Not only we use it extensively internally, we are also embedding GraphQL interfaces in some of our products, we have open source packages like ibm/OpenAPI-to-GraphQL and IBM is a founding member of the GraphQL Foundation.


caniszczyk:

graphql/graphql-landscape#48

PR #759, opened 2019-08-20
Update footer

JoelMarcey:


caniszczyk:
Thanks for the catch @JoelMarcey

JoelMarcey:
@caniszczyk I think it was caught by someone on Twitter :)

PR #764, opened 2019-09-01
Simplify the apollo-server example

lewislbr:
Reduced the complexity of explaining the apollo-server implementation with a Node.js framework, when it can be implemented as stand-alone server, by simplifying the example code and rewriting the description following the official Apollo documentation.


orta:
This was updated to their new example code (which is simpler 👍  )

PR #770, opened 2019-09-18
Upd docs: Graphql with JavaScript Express server

Ajmdag:
Previous version of docs worked only with apollo-server-express 1.x.x.

Now available 2.9.3 (https://github.com/apollographql/apollo-server/tree/master/packages/apollo-server-express)


orta:
Cool, thanks, looks like the docs they recommend 👍

PR #782, opened 2019-10-09
Berlin Conf date has passed and is no longer upcoming

M0nica:
- I updated to remove berlin from the upcoming conf list as it is no longer upcoming. Should historical conference information live somewhere?

### DIFF
--- a/site/community/Community-Events.md
+++ b/site/community/Community-Events.md
@@ -8,14 +8,6 @@ sublinks: Upcoming Events,Meetups
 
 ## Upcoming Events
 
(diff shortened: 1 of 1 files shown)


orta:
Great question, I wonder if they should if there's a youtube playlist links?

brianwarner:
@orta yes, here you go: https://www.youtube.com/watch?v=e5xaIRUMtjs&list=PLyspMSh4XhLMAIqlh3Z5R6frHMDc7t3eG

orta:
Thanks

PR #787, opened 2019-10-19
add Diana.jl

codeneomatrix:


caniszczyk:
the implementation looks incomplete, so closing

PR #788, opened 2019-10-20
swap languages to get proper alphabetical order

hedgesky:
This PR changes order of languages a bit to make it sorted. The rest of the list is alphabetically ordered but Ruby (presumably incorrectly) goes after Scala. This also conforms with the order the libraries descriptions go below.


orta:
Thanks 👍

PR #795, opened 2019-11-15
Add juniper Rust for server libraries

smyrick:
juniper is currently on of the most active and supported libraries for running a GraphQL server in Rust.

https://github.com/graphql-rust/juniper


orta:
👍

PR #800, opened 2019-11-27
New swapi-graphql service using netlify functions

acao:
Now we can merge this PR, and https://swapi-graphql.netlify.com is live 🌍 at graphiql.org/swapi-graphql :D

here's where you can learn about the new features:
https://github.com/graphql/swapi-graphql/pull/162

I'm pretty sure the order of the last two rules will be fine. If the last one doesn't work, it will still work

```
/swapi-graphql/graphql https://swapi-graphql.netlify.com/.netlify/functions/index 200
```
this just 200 redirects (non permanent?) which essentially is netlify terms for what nginx is a proxy_path

so this will just make `graphiql.org/swapi-graphql/graphql` an alias for the lambda endpoint. see the PR's develop preview before it's merged if you want to see that default lambda path in action for yourself

PR #803, opened 2019-11-29
Revert "New swapi-graphql service using netlify functions"

acao:
Reverts graphql/graphql.github.io#800

This worked locally with `netlify dev` but the redirect doesn't seem to be working at all


orta:
👍

PR #805, opened 2019-12-04
add graphql asia in upcoming events

praveenweb:
GraphQL Summit is over. So I'm adding GraphQL Asia conference under upcoming events.

I'm also the co-organiser of GraphQL Bangalore meetup. I have updated the link to avoid redirects.


orta:
👍

PR #807, opened 2019-12-04
remove / update dead links

praveenweb:
This PR removes and updates links that are not working, not active anymore or redirecting to a different URL.

Here are some notes with reasoning.
- Scaphold BaaS isn't active anymore. The link doesn't work and their open source code isn't updated anymore.
- Reindex has [discontinued their BaaS](https://www.reindex.io/blog/discontinuing-backend-as-a-service/) from 2017, the link gives a 404 and they are not working on the open source code.
- Removed their twitter / slack links which aren't active as well
- Updated Apollo blog link
- graph.cool/blog redirects.
- learngraphql.com doesn't work
- graphql-world.com doesn't work
- graph.cool/talks doesn't work
- learnapollo.com & learnrelay.org both redirect to howtographql.com which is already listed at the top.


orta:
Nice, thanks!

PR #802, opened 2019-11-29
Update community Resources

packtpartner:
Added a new request

PR #813, opened 2019-12-30
update Community resources

packtpartner:
Added new request to include newly published video titles


orta:
Hi, the same feedback from https://github.com/graphql/graphql.github.io/pull/802 applies - if you remove the tracking params from the URLs they should be ok

orta:
poke

PR #815, opened 2020-01-17
add dynamic year to footer

ahmdtalaat:


orta:
Good call, let's get that merged 👍

Issue #835, opened 2020-02-22
Fix CodeMirror ScrollBars


elit-altum:
__Issue__
An attempt has been made in the codebase to hide the vertical and horizontal scrollbars in CodeMirror for the GraphQL code examples and returned queries. This has not been successful and should be fixed.

__Screenshots__

![Issue-Scrollbars](https://user-images.githubusercontent.com/41413622/75095933-f70a8e00-55bf-11ea-8d1e-38ae5bb384ec.gif)

__How to recreate the issue__
- Go to any page where code examples are rendered using CodeMirror (eg: https://graphql.org/learn/queries/ )
- See the scrollbar on every code example with overflow content

__Fixing__
I would like to work and fix the issue.

PR #836, opened 2020-02-22
Fixes CodeMirror ScrollBars

elit-altum:
Fixes:  #835 

The CodeMirror scrollbars have been hidden without tampering with the scroll functionality. These provide a more cleaner and focused UI. 
The code has been tested on the current versions of Firefox, Edge and Chrome.

__Screenshots__
Fixed-Scrollbars

### DIFF
--- a/site/_css/codemirror.less
+++ b/site/_css/codemirror.less
@@ -178,15 +178,8 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket {color: #f22;}
   z-index: 6;
   display: none;
 }
-.CodeMirror-vscrollbar {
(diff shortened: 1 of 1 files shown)


orta:
cool, yeah, this looks like the intention of the original changes 👍

shashwatjolly:
Shouldn't there be some indication that the section is scrollable?

elit-altum:
> 
> 
> Shouldn't there be some indication that the section is scrollable?

The scrollable sections have some content slightly hidden by the borders. Along with that the query examples are very intuitive to indicate that there is more data.

PR #839, opened 2020-02-23
graphql logo position fix on small screens

RoLoDeXx:
<b>Current Behavior</b>

Screenshot from 2020-02-23 00-37-16

<b>Expected Behavior</b>
Screenshot from 2020-02-23 00-38-44


carolstran:
Pinging @IvanGoncharov - this is the same fix as #850, so I guess it depends on which breakpoint/CSS unit you prefer.

PR #840, opened 2020-02-23
Fix #831 - Search bar is getting out of the navbar when selected

### LINKED ISSUE #831: search bar error
### Screenshot:
Screenshot from 2020-02-21 19-10-54

### Additional context
if really a bug,i would like to take this up.

ani4aniket:
Fix #831 
Screenshot 2020-02-23 at 3 37 35 PM

### DIFF
--- a/site/_css/algolia.less
+++ b/site/_css/algolia.less
@@ -27,7 +27,7 @@
     transition: width 0.2s ease-out;
 
     &:focus {
-      width: 300px;
+      width: 250px;
     }
   }
 


site/_css/algolia.less
```diff
@@ -27,7 +27,7 @@
     transition: width 0.2s ease-out;
 
     &:focus {
-      width: 300px;
+      width: 250px;
```
elit-altum: This much difference will then move it to the bottom on a smaller screen resolution. This may work on your screen size but will not in many others. This is not a permanent fix or solution.

ani4aniket: Actually we need a sidebar rather than navbar for smaller devices.

carolstran:
👋🏼 @ani4aniket, thanks for taking the time to make a pull request! As there are a couple PRs for this issue open, I'm going to close this one in favor of #841 as it incorporates media queries. I'd be super curious to hear your review of that PR if you have time 😊

PR #844, opened 2020-02-25
Update Learn-Queries.md

leventov:
In sentence

> When something goes wrong either in your network *logs* or your GraphQL server, it is easier to identify a query in your codebase by name instead of trying to decipher the contents.

Removed the word "logs" because it looks like it doesn't belong to this sentence.


orta:
Thanks for the PR and helping to improve the docs, but I think it makes more sense with the word logs in. 

- With: If you have problems look at your network logs: see what the response to your query is

- Without: If you have problems look at your network: check computer is connected to the net

leventov:
I see what was meant now, but it's not obvious meaning. Suggested the following:

"When something goes wrong (you see errors either in your network logs, or the logs of your GraphQL server) it is easier to identify a query in your codebase by name instead of trying to decipher the contents."

@orta WDYT?

orta:
Yep :+1:

PR #845, opened 2020-02-25
Update Learn-Schema.md

leventov:
> But it's useful to have an exact description of the data we can ask for - what fields can we select? What kinds of objects might they *return*?

In this section, in the second sentence, "they" refers to "fields" so it doesn't seem right to say "field returns a type", so replaced with *contain*.


hwillson:
Thanks @leventov, I agree 👍 - would you mind signing the new CLA when you have a chance? Thanks again!

saihaj:
This PR doesn't seem to have been touched in a considerable amount of time. Additionally, we have implemented changes to the learning sections, which have also generated conflicts that require resolution. Closing this for now. Feel free to create a new PR.

PR #847, opened 2020-02-27
Adding Graphaello to the list of iOS Libraries

nerdsupremacist:
While I was interning at Facebook last year, I learned how useful Relay could be when building React Apps that use GraphQL.

When SwiftUI came out, I just had to make something similar for Swift. And took some more liberties to make it Swiftier.
Hope you guys like it!

Here's the link again: https://github.com/nerdsupremacist/Graphaello


orta:
Cool!

PR #861, opened 2020-04-14
Add async-graphql rust crate

sunli829:
Repo: https://github.com/sunli829/async-graphql
Examples: https://github.com/sunli829/async-graphql-examples

It supports all of the GraphQL specifications and is easy to integrate into existing web servers.


phated:
Please add this! It's an amazing library that solves a lot of the edge cases people might encounter with Juniper.

sunli829:
This seems like a pretty good design, but unfortunately it cannot be implemented due to the limitations of the rust procedural macros. The current interface is defined in a way that I've thought about for a long time, is easy to use and easy to understand, and more importantly is rustfmt friendly. See how juniper defines the interface. It's similar to what you want, but not easy to use.

PR #862, opened 2020-04-21
Change 'graphql-server' to 'apollo-server'

jacobdcastro:
It seems Apollo has updated the name of the package for their GraphQL server to `apollo-server`. Just wanted to make it more accurate.


carolstran (approved):
Great catch @jacobdcastro! The link redirects correctly, but I agree that it'd be better to have the naming accurate in the docs 💯 

Pinging @IvanGoncharov to merge

PR #887, opened 2020-07-06
Fix import of "graphqlHTTP" (Tutorial-ExpressGraphQL)

domenic-corso:
Need to use object destructuring here as the "express-graphql" package gives us an object containing property "graphqlHTTP"


carolstran (approved):
Thanks for taking care @domenic-corso 🍨 

Pinging @IvanGoncharov to merge - similar to #899 and #905, but covers a different piece of example code. It could be worth searching through the source code once these PRs are merged to check for any other instances.

PR #891, opened 2020-07-08
update index.html.js

Ashish1101:
var graphqlHTTP = require('express-graphl') throwing an error .
graphqlHTTP is not a function 
but destructuring it from express-graphql solves the problem

var { graphqlHTTP } = require('express-graphql')


Ashish1101:
please check.
This small problem in the documentation took my 3 hours

PR #894, opened 2020-07-14
Update link of serialization format in spec

outloudvi:


carolstran (approved):
Great catch @outloudvi 🎯 

Pinging @IvanGoncharov to merge!

site/learn/BestPractice-Introduction.md
```diff
@@ -23,7 +23,7 @@ Read more about this in [Serving over HTTP](/learn/serving-over-http/).
 
 ### JSON (with GZIP)
 
-GraphQL services typically respond using JSON, however the GraphQL spec [does not require it](http://facebook.github.io/graphql/#sec-Serialization-Format). JSON may seem like an odd choice for an API layer promising better network performance, however because it is mostly text, it compresses exceptionally well with GZIP.
+GraphQL services typically respond using JSON, however the GraphQL spec [does not require it](http://spec.graphql.org/June2018/#sec-Serialization-Format). JSON may seem like an odd choice for an API layer promising better network performance, however because it is mostly text, it compresses exceptionally well with GZIP.
```
IvanGoncharov, line 26: Better to use a draft since this way we don't need to update after every release.
```suggestion
GraphQL services typically respond using JSON, however the GraphQL spec [does not require it](http://spec.graphql.org/draft/#sec-Serialization-Format). JSON may seem like an odd choice for an API layer promising better network performance, however because it is mostly text, it compresses exceptionally well with GZIP.
```

Issue #895, opened 2020-07-15
Sample code is incorrect (generates errors)


jbrown123:
The sample code on several pages is incorrect (generates runtime errors).
```
app.use('/graphql', graphqlHTTP({
                    ^

TypeError: graphqlHTTP is not a function
```

The following pages are affected:
* Passing Arguments
* Object Types
* Mutations and input types

You left out the brackets around the graphqlHTTP variable.

You have
```
var graphqlHTTP = require('express-graphql');
```

Should be
```
var { graphqlHTTP } = require('express-graphql');
```

If I find more instances of this same mistake, I'll update this issue.

IvanGoncharov:
Fixed in #899

PR #899, opened 2020-07-26
Fix import of express-graphql

RitikDua:
Fixed a syntax error during the destruction of function "graphqlHTTP".

### DIFF
--- a/site/graphql-js/Tutorial-Authentication.md
+++ b/site/graphql-js/Tutorial-Authentication.md
@@ -15,7 +15,7 @@ For example, let's say we wanted our server to log the IP address of every reque
 
 ```javascript
 var express = require('express');
-var graphqlHTTP = require('express-graphql');
+var {graphqlHTTP} = require('express-graphql');
 var { buildSchema } = require('graphql');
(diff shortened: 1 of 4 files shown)


carolstran (approved):
Thanks for taking care @RitikDua 💫 

My only suggestion would be to add spaces between the graphqlHTTP and the brackets (ie `{ graphqlHTTP }` instead of the existing `{graphqlHTTP}`) to keep the style consistent with the other imports.

Pinging @IvanGoncharov - this PR is similar to #905, but updates more instances of the import error. Also closes #895.

RitikDua:
> Thanks for taking care @RitikDua dizzy
> 
> My only suggestion would be to add spaces between the graphqlHTTP and the brackets (ie `{ graphqlHTTP }` instead of the existing `{graphqlHTTP}`) to keep the style consistent with the other imports.
> 
> Pinging @IvanGoncharov - this PR is similar to #905, but updates more instances of the import error. Also closes #895.

Thanks @carolstran  for reviewing it and i also completed the changes you sugggested,.

PR #900, opened 2020-07-27
Litte changes on the doc

sadiqful:
I edited some part of the doc that had "We launched 2 years ago" and other minor improvements.


carolstran:
👋🏼 The decision is ultimately up to @orta, but I don't think it makes sense to update the blog posts as you proposed. The posts are dated, so it's clear that (for example) graphql.org was published one year before the date of the post. Plus since it's a blog post and not official documentation, it doesn't need to be maintained in the same way.

Happy to hear counterpoints if anyone disagrees! cc: @IvanGoncharov

PR #905, opened 2020-08-13
Fix import error in GraphQL.js sample code

AmmonMatthews:
changed graphqlHTTP to { graphqlHTTP }. It was erroring out and doesn't look like it is a default export but a named export from the error I received saying TypeError: graphqlHTTP is not a function


carolstran (approved):
Looks good, thanks for taking care of this @AmmonMatthews! Also closes #895. 

My only note would be to update the commit message to give a bit more context (ie. `Fix import error in GraphQL.js sample code` instead of `Changed line 21`) - but that could happen when merging.

Pinging @IvanGoncharov to merge 😊

Issue #790, opened 2019-10-21
High CPU usage on homepage


benlesh:
I left https://graphql.org/ open on another screen and noticed my MBP's fan spin up. I was curious to see what what eating my CPU, and when I looked it was Google Chrome Renderer. Upon checking the Chrome Task Manager, I found that the site was using 30-40% CPU, which seems excessive. When I dug in, it looks like there are just some JS animations that are eating most of the CPU. Looking at the animations, it seems like they could all just be CSS animations and that would help the page perform better, overall.

Just wanted to bring it to your attention. Otherwise, it's a great site.

elit-altum:
Hey @benlesh ! 
Thank you for bringing this up. Could you elaborate upon which particular animations you found were causing this issue, so I can work around it ?
Any help from your side would be really useful. Thank You!

saihaj:
If we are going to keep the animations same in new iterations (when Gatsby site is in prod) I suggest that we solve this since it fixes one of the issues mentioned in #896 @IvanGoncharov @carolstran

carolstran:
Yup totally agree @saihaj and thank you @benlesh for flagging this!

IvanGoncharov:
@saihaj @carolstran I personally experience this bug from time to time and it drives me crazy so definitely something we should address.

If someone wants to work on this please claim this issue in comments.

JanetLaysEggs:
Well one solution (may seem a bit naff) is to cheat and use event driven images, gifs, and videos for the animations (as lovely as the JS animations are) instead as they should be less CPU intensive? Websites like apple.com and tunnelbear.com do it all the time. A really good use case of this would be the text going forwards and back between the colourful servers and the phone on the main page as currently a lot of text is having to bounce up and down the page with lots of colour changes.

However there is an argument in how long the viewers are actually on the main page where the animations live... I doubt normal viewers checking out what GraphQL is etc are going to hang around on the index page for more than 10-20 seconds - after which the new pages they navigate to will load and they won't experience the effects of the large CPU usage like the whirring of fans.

ardatan:
Is this issue still relevant after Gatsby migration?

saihaj:
Yes it is still relevant after the migration since the issue is caused because they are JS based animations

Urigo:
This should be solved now thanks to @dimaMachina's refactor of using Nextra
Please reopen in case that's not the case

PR #913, opened 2020-09-04
Migrate to Gatsby 🚀

IvanGoncharov:
Based on all the great work done by @saihaj in #896 👍 

@saihaj You did an excellent job describing missing pieces in #896 
Can you please post it here so I can pin this comment?
I can also assign this PR to you if you don't mind and you would be able to coordinate other people's help.
Just ping me in Slack if you need anything merged into the Gatsby branch.


IvanGoncharov:
One issue: I accidentally squashed all your commits into one. I will redo this branch so we can preserve the commit history. It shouldn't affect this PR through.

saihaj:
### Remaining Issues 

* [ ] Fix DOM nesting errors: CSS needs to be modified so that instead of forcing a body tag inside a div `components/Layout.js` a div with some className is used
* [ ] Fix animations in `pages/index.js`: `useEffect` hook is used to run the animations but the problem is when page changes it still runs for some reason and I don't know how to fix it. Update: I tried a clean up function to remove the child nodes but no good.
* [ ] Code highlighting: I was able to prefix classnames where needed for code highlighting but some colors are not working because there are different classnames when compared to `css/Prism.less`
* [ ] Show title of next page in "continue reading" button. I tried passing the permalink to context from `gatsby-node.js` so that I can run one more query that will return title and one that returns content . When running it in GraphiQL (single query) the results are as expected but when I combine it with page query then the title is not correct.
* [ ] CNCF iFrame -> I was able to add landscape but don't couldn't get the resizer script working
* [ ] Add [Algolia search plugin](https://www.gatsbyjs.org/docs/adding-search-with-algolia/#setting-up-the-project) (Need maintainer to set this up) 
* [ ] [Google Analytics](https://www.gatsbyjs.org/packages/gatsby-plugin-google-analytics/) setup (Need maintainer to set this up)

Since Algolia and GA keys will go `.env` and need to set those accordingly in CI pipeline too

#### Possibly can refactor but don't know how to: 
* Since I can't pass in variables to static query (https://github.com/gatsbyjs/gatsby/issues/10482) there was not an elegant way other than separating the logic in a separate component (`components/Sidebar/Sidebar.js`) and pass down data as props from a separate component that runs static query.
* I think it is possible to do same thing for `templates`

> I can also assign this PR to you if you don't mind and you would be able to coordinate other people's help.
👍🏼 Yes please

saihaj:
>  Fix animations in pages/index.js: useEffect hook is used to run the animations but the problem is when page changes it still runs for some reason and I don't know how to fix it. Update: I tried a clean up function to remove the child nodes but no good.

#790 can be resolved by this

carolstran:
Should we close this because #938 was merged?

saihaj:
Yes

carolstran:
Alright closing then!

PR #917, opened 2020-09-18
Fix wrong meaning in "thinking in graphs - legacy"

0000marcell:
I think the "how" and "what" should be swapped here since it doesn't
make much sense based on the previous phrase.


carolstran (approved):
👋🏼  @0000marcell, thanks for taking the time to open a pull request! I agree that this is confusing and I would also assume that these two should be switched.

Pinging @IvanGoncharov in case I'm wrong - also since he'll be the one to merge once we're accepting content changes again.

carolstran:
Hey hey 👋🏼  We recently completed a migration to Gatsby (#938), so could you please resolve the conflicts on your branch? Then I can merge.

0000marcell:
I think it's everything ok now :+1:

PR #920, opened 2020-09-26
Change duplicate body tag to a div

mikeesto:
Resolves

`Fix DOM nesting errors: CSS needs to be modified so that instead of forcing a body tag inside a div components/Layout.js a div with some className is used`

in https://github.com/graphql/graphql.github.io/pull/913#issuecomment-687325610


saihaj:
Ping @IvanGoncharov @carolstran

IvanGoncharov:
@mikeesto Thanks 👍

PR #918, opened 2020-09-25
Fix "TypeError: graphqlHTTP is not a function" in Passing Arguments page, GraphQL js tutorial

echamudi:
The current example of the rollDice API code contains an error.
```
TypeError: graphqlHTTP is not a function
```

This patch will make it run.

PR #925, opened 2020-10-06
Update import statement in "Passing Arguments" from Tutorial-PassingArguments.md 

joshwreford:
The import statement for `graphqlHTTP` is incorrect. It throws an error when copied and pasted locally.


carolstran:
Good fix, but as you already mentioned - this is a duplicate of #918. So I'm going to close it, but thanks for caring enough to open a fix!

PR #931, opened 2020-10-15
Update Tutorial-PassingArguments.md

muzhaqi16:
use named import when importing graphqlHTTP from express-graphql
Added curly brackets


carolstran:
Good fix 🐛 🔥  This is a duplicate of #918 though, so I'm going to close it - but thanks for caring enough to open a fix!

PR #933, opened 2020-10-15
Fix graphqlHTTP imports

carolstran:
There have been a bunch of PRs related to the lack of curly braces when importing `graphqlHTTP` in code samples - so I went through and fixed the rest of the bugs 🧑‍🚒


IvanGoncharov (approved):
@carolstran Thanks a lot 👍

PR #934, opened 2020-10-17
Clean up useEffect animations (Gatsby)

mikeesto:
Resolves:

>Fix animations in pages/index.js: useEffect hook is used to run the animations but the problem is when page changes it still runs for some reason and I don't know how to fix it. Update: I tried a clean up function to remove the child nodes but no good.

In https://github.com/graphql/graphql.github.io/pull/913#issuecomment-687325610

There's some good discussion in https://github.com/graphql/graphql.github.io/issues/790 about using CSS or GIFs/videos for the animations. This is more a bandaid fix!

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -88,6 +88,7 @@ const IndexPage = () => {
   useEffect( () => {
     let i = 0
     let forward = true
+    let timer
 
     const showResponse = num => {
       document.getElementById( 'r1' ).style.display = num === 1 ? 'block' : 'none'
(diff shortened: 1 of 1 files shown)


carolstran (approved):
I'm fine with this as a temporary fix! Especially as we have an issue open about CSS animations already. Pinging @IvanGoncharov to review before merging 🕺🏼 

Thank you @mikeesto!

Urigo:
Thank you @mikeesto !
We've integrated it into https://github.com/graphql/graphql.github.io/pull/936
Could you review and make sure it work good there?

Thanks

mikeesto:
Thanks @Urigo . This has been merged in https://github.com/graphql/graphql.github.io/pull/938 so I'm closing this.

PR #937, opened 2020-10-20
Fix fragment name

doubledup:
Conditional fragments aren't referenced anywhere else in the documentation,
going by results from the search bar. This is introduced as an inline fragment
in the "Queries and Mutations" section.


carolstran (approved):
Great catch! I definitely agree that we should be using consistent terms to avoid confusion. Pinging @IvanGoncharov in case I'm missing something, but otherwise, I think this looks good to merge 💫 

Thank you @doubledup 💐

carolstran:
Hey hey 👋🏼  We recently completed a Gatsby migration (#938), so could you please resolve the conflicts on your branch? Then I can merge.

PR #939, opened 2020-11-02
Fix search for gatsby site

orta:
Roughly: The previous version of the site would load the page from scratch every time, so calling an external lib at the footer was a reasonable call. Gatsby on the other hand, won't do a full reliable re-render of a page and so when you're interacting with `script` tags etc. There needs to be a bit more dancing to hook stuff up.

This technique is almost C&P'd from the TypeScript website. A different potential fix is to move to the v3 docsearch implementation - https://github.com/microsoft/TypeScript-Website/pull/928, which is done entirely within the bundler context and algolia have done a good job making it work in theses SPA-y sites.

For now this is enough to make it all work right

### DIFF
--- a/src/components/Footer/index.tsx
+++ b/src/components/Footer/index.tsx
@@ -108,21 +108,6 @@ const Footer = ({sourcePath}: { sourcePath: string }) => {
         </section>
       </footer>
 
-      <script
-        type="text/javascript"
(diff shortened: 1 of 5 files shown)


orta:
Screen Shot 2020-11-02 at 10 04 29 AM

PR #938, opened 2020-10-21
Gatsby Migration

ardatan:
Hi everyone.
[We](https://the-guild.dev) @the-guild-org have decided to take on the work of the Gatsby migration in order to unblock all the contributions for graphql.org.
We tried to start with the amazing work that has been done in https://github.com/graphql/graphql.github.io/pull/913 by @saihaj but decided it was actually faster for us to start from scratch and finish all the things that need to be done in order to get this work merged.
We hope it will be ok with you to move the work here until we get it done.
If you disagree we are open to hear and move back to the former PR.

The main differences from the former PR are mostly around styling and making the migration pixel perfect as the current website.
Here are some differences we noticed on the former PR and made sure on this one:

- Make live MiniGraphiQL code examples work
- Made code highlighting work exactly as the existing website
- Made the “How’s using GraphQL” at the bottom of the first page pixel perfect like the existing website
- Got back the “edit this page” button
- Got the the GraphQL Logo at the bottom of the first page
- Sidebar on “Learn” section is now again pixel perfect as the existing website
- Pixel perfect buttons on the “Learn” section
- Same styling for highlighted text
- Show title of next page in "continue reading" button

Those styling errors are obvious, just open those 3 links and move between each other:

- Old PR: https://existing-gatsby-graphql-org.netlify.app/
- Current website: https://graphql.org/
- New PR: https://guild-gatsby-graphql-org.netlify.app/

Also, we fixed all the remaining issues from PR https://github.com/graphql/graphql.github.io/pull/913 (some just solved themselves with the new code) and backported all the fixes and PRs that were added on top of it.

We’ve also upgraded all dependencies to latest (including all the old graphql related dependencies).
(except codemirror-graphql - we’ve found an issue there and submitted a PR, @acao let me know if you have time to go through it together quickly. anyway it’s not a blocker as the former version works)

Here is a link to an up to date deployment of the branch: https://guild-gatsby-graphql-org.netlify.app

We believe this PR is in a state that is better then the current deployed version and ready to be merged.
@saihaj @leebyron @IvanGoncharov @carolstran @mikeesto please review it and if there is nothing that is worse then the current website, let’s merge and deploy the new version asap and open again the options for people to contribute!

Last thing, in order to deploy this branch we need the credentials for Netfliy (to add GitHub token for our new sorting algorithm), Algolia (to have better indexing using gatsby plugin) and Google Analytics - @leebyron @brianwarner @caniszczyk could you help us with that?

PR #940, opened 2020-11-02
Update dependencies

ardatan:


saihaj (changes requested):
We can use the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#gatsby). React team provides a codemod to remove all `import React from ‘react’`

orta:
Do you want to do the transformed @ardatan? I don't end either way

ardatan:
I was afraid of breaking things. Let's leave it in a different PR.

saihaj:
We could have used this new transform syntax in #938 since that was a feature added in v16.14 but I missed to point that out in review. 

According to [release blog post](https://reactjs.org/blog/2020/10/20/react-v17.html#react). They changed event delegation and a few other non breaking issues. I don’t think any of the changes will affect us. Purpose of this version is to make it easier to gradually upgrade react’s future versions.

Sure we can get that in a new PR

Urigo:
I've reviewed and everything works as expected so I'm merging, but that doesn't mean we shouldn't move to the new JSX transform as @saihaj suggested!
I suggest we'll try from now on to keep PRs small and easy to merge, so we could move more quickly.

@saihaj that's a good idea, are you up to creating a new PR with the new transform?

saihaj:
I can get something in today

saihaj:
@Urigo turns out gatsby has disabled automatic JSX runtime temporarily https://github.com/gatsbyjs/gatsby/pull/27615. We need to wait for a new Gatsby release before we refactor.

Issue #558, opened 2018-08-03
Menu on the right could be made sticky


mishasaggi:
For a better experience.

carolstran:
👋🏼 @mishasaggi, thanks so much for opening an issue and a PR on the GraphQL docs! I swear it's appreciated even if it's been two years since you opened it 😅

Something I wonder about this proposal (and after testing your PR) is if it would really provide a better experience? I agree that not having the sidebar menu available once you scroll past it isn't great, but I wonder if having to scroll all the way to the end of the article to reach the bottom of menu (at least when you're on a standard laptop screen) is a better experience... 🤔 

Genuinely would be curious to hear your thoughts and reasoning!

PR #953, opened 2020-11-04
make sticky side menu items

igbominadeveloper:
#### What does this PR do?
This PR adds styles to make the sidebar menu sticky on desktop screens

#### Background context
The sidenav keeps scrolling with the page and makes it difficult when users need to navigate forcing them to scroll back up before they can navigate

#### What are relevant issue(s)?
#558 

#### Screenshots
Annotation 2020-005-09 215804

Annotation 2020-05-09 215804

### DIFF
--- a/src/assets/css/_css/docs.less
+++ b/src/assets/css/_css/docs.less
@@ -248,6 +248,10 @@
   margin: 1em 0 4em;
   width: 195px;
   padding-left: 20px;
+  position: -webkit-sticky;
+    position: sticky;
+    top: 24px;
+    height: 100%;
 
   @media screen and (max-width: 740px) {
     width: auto;


orta:
Continuing the feedback in #558 - I'm not sure we should do this either, but this can't go in as it is because it means you have to scroll to the bottom in order to hit some of the links in the sidebar. 

Adding a scrollview for the sidebar is an answer if it vertically overflows and allows scrolling inside the sidebar, but the permanent  scrollbars feels bad on windows/linux too. So I'm leaning towards no.

igbominadeveloper:
> Continuing the feedback in #558 - I'm not sure we should do this either, but this can't go in as it is because it means you have to scroll to the bottom in order to hit some of the links in the sidebar.
> 
> Adding a scrollview for the sidebar is an answer if it vertically overflows and allows scrolling inside the sidebar, but the permanent scrollbars feels bad on windows/linux too. So I'm leaning towards no.

I understand this now and agree it is a bad UX to have to scroll to the bottom before you access some links. Although, I like the idea of adding overflow to the sidebar, I am not certain it would look great. Nevertheless, I shall give it a try

Urigo:
please open a new PR when you have the right UI improvement in mind
Thank you!

PR #969, opened 2020-11-19
Update documentation for urql

kitten:
I've noticed that the entry for `urql` is actually rather old from back when our bindings were pretty much React exclusive, which isn't the case anymore, so I've added some more notes to bring all the taglines in line (while also making sure that it's consistent across our sites & docs https://github.com/FormidableLabs/urql/pull/1177)

I've seen that other libraries expand beyond their shorter description with some more content, so I've shortened and added some content to the `urql.md` page from our own readme.

The links have been updated to reflect the core packages and go to the documentation directly.

Cheers! 🙌


carolstran (approved):
Sweeeeeet 🥳  

Thanks for taking the time to update this!

orta:
Perfect

Issue #946, opened 2020-11-03
Building schema error when running project with NPM


mikeesto:
When trying to `npm install && npm start`:

<img width="809" alt="Screen Shot 2020-11-03 at 2 57 30 pm" src="https://user-images.githubusercontent.com/21051488/97952005-c124d700-1de7-11eb-9622-1e360b19cf2c.png">

There seems to be some related discussion about it [here](https://github.com/gatsbyjs/gatsby/issues/13278) although there's no resolution. I ended up deleting node_modules & package-lock.json, and using yarn.

carolstran:
@mikeesto I also ran into this the first time I launched the site post-Gatsby migration. I'm working on a new contributing guide and thinking of only encouraging the use of `yarn` for now since there's no resolution on the Gatsby end 🤷🏼‍♀️

orta:
Yep, we should drop the npm installation recommendations - they don't get the lockfile and will definitely have different builds which would cause random issues like this

Urigo:
@TuvalSimha is this still relevant?

TuvalSimha:
@Urigo 
I was trying to reproduce it:
1. `npm install`
2. `npm start`

The build was created and the project was running properly. You can close this issue for this moment.

Urigo:
thanks!
solved, please reopen if happens again

PR #970, opened 2020-11-19
Add contributing guide

carolstran:
We've discussed the idea of updating the internal documentation in multiple places (most recently #946), so this PR is a start to this 💫 

Next up after this would be updating the README because this PR makes it extra... scarce...


orta (approved):
Nice one, looks good to me 👍🏻

PR #976, opened 2020-11-23
Re-add NGraphQL link in Code page (for .NET)

rivantsov:


carolstran:
@orta can correct me if I'm wrong, but I believe with the new Code page we'd prefer if you write a short description or code sample when adding a new library to the list 😊

rivantsov:
thank you. I was thinking about adding code snippet, but unfortunately, by design (and .NET nature), the smallest possible setup involves about half a dozen classes. NGraphQL does not support schema-first, only code first (by design), and this requires more than a few lines. I will add code snippet later to client library reference. 
I suggest you guys do not focus too much on code snippets. Not all languages are like JavaScript and offer a trivial sample. - which is not necessarily disadvantage. 
To expand on description - I don't know, the essence is there in this sentence - library for building GraphQL servers. All I can add is marketing fluff. 
Also guys, please do not put too much weight on this description. In general, no serious developer will make a choice based on just reading this short description or snippets on your page. He/she will inspect ALL links in details, downloading the code and spending quite a lot of time on each solution. The description on your site could only impact the order in which he'd go thru the list, but not the process or final decision. 
thank you.

rivantsov:
I added a few sentences to description. 
thank you

orta:
Yeah, we'd prefer it to be enough so that someone doesn't have to click through to make a decision. Showing the API usage isn't really marketing fluff IMO. Given you have opinions against it, I don't see it as too much of a blocker though :+1:

rivantsov:
thanks so much, really appreciate it!

PR #987, opened 2020-12-10
Update contributing guide to include browser support

carolstran:
Based on a discussion in #985. 

I thought it'd be nice to note what browsers we aim to support in the contributing guide so that people know what to test when they are making code or design changes to the website 🔨 

Kept it vague for now, but if we want to mention specific versions - let me know and I'll update.


orta:
Cool, looks good 👍🏻

PR #990, opened 2020-12-12
Add graphql-hooks client library

simoneb:


orta:
Hi, it'd be great if you added some markdown to that file describing how to use the API and give folks a reason to dig deeper

simoneb:
> Hi, it'd be great if you added some markdown to that file describing how to use the API and give folks a reason to dig deeper

I've extended the description a bit

orta:
Sure, you're welcome to skip the markdown

simoneb:
> Sure, you're welcome to skip the markdown

I didn't mean to reject your advice, I simply kept it in line with the other client libraries listed in that same page, which are not providing more than a description, unless I missed something.

orta:
Yeah, I'm not offended 👍🏻 

The markdown section is new, and we've not been actively going back to add them to older libs. Showing their API usage just makes it easier for folks to make informed decisions without a lot of clicks.

simoneb:
I only now realized what you meant. I missed it because none of the client libraries in the same group provides examples, but the servers do. I did it in https://github.com/graphql/graphql.github.io/pull/991. Thanks for bringing that you.

PR #991, opened 2020-12-15
Add graphql-hooks example

simoneb:
## Description

Includes example of how to use the library, as suggested in https://github.com/graphql/graphql.github.io/pull/990#issuecomment-744055839


orta:
Thanks, looks great

PR #981, opened 2020-11-26
Added "GraphQL by PoP" server in PHP

leoloso:
Added [GraphQL by PoP](https://graphql-by-pop.com/) to the list of GraphQL servers in PHP

PR #1000, opened 2021-01-14
Updated GraphQL API for WordPress website URL

leoloso:
## Description

Update server's website to its new site https://graphql-api.com

### DIFF
--- a/src/content/code/language-support/php/server/graphql-api-for-wordpress.md
+++ b/src/content/code/language-support/php/server/graphql-api-for-wordpress.md
@@ -1,7 +1,7 @@
 ---
 name: GraphQL API for WordPress
 description: A GraphQL server for WordPress
-url: https://github.com/GraphQLAPI/graphql-api-for-wp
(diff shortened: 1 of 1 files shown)


orta:
👍🏻 cool, looks good to me

leoloso:
Thanks for merging @orta. Any chance to also merge #981? Because the description for the GraphQL API for WordPress reads "Based on GraphQL by PoP", which is why I didn't add implementation details here, since they were already added to this other server

orta:
Great point, I missed that!

leoloso:
Lovely, thanks!

GraphQL by PoP is still not in the list though, and it's been over 1 hour since merged. Does it take long to deploy the site? Or if not, maybe there's some issue going on?

https://graphql.org/code/#php

If there's an issue, it may be because both GraphQL by PoP and GraphQL API for WordPress have leoloso/PoP as repo... (it's the monorepo containing both projects)

leoloso:
It's there now, great!

PR #1012, opened 2021-02-27
Fix links that have renamed their branch to main

stingalleman:
Not really anything important, they now just redirect to the correct link.


orta:
Thanks!

PR #1023, opened 2021-03-12
Add Ferry client

agent3bood:
Closes #<issue number>

## Description

Mention the Ferry GraphQL client on this guide 
https://github.com/gql-dart/ferry


orta:
Do you have a reason for not including an example of the API in the markdown?

agent3bood:
> Looks good - but do you have a reason for not including an example of the API in the markdown?

There are different ways to use that package, I looked but its not easy to find a sample code that sum it all. 

The library website will walk the user through setting the client and using it.

orta:
OK 👍🏻

PR #1025, opened 2021-03-18
Redirect graphql.github.io pages to their counterparts on graphql.org

brianwarner:
This PR is a proposed solution to the fact that we currently have two active sites, graphql.org (the canonical site) and graphql.github.io (the preceding version). This PR replaces all content files on the GitHub Pages-based site with redirects to their matching pages on the canonical graphql.org site. All page pre-prereqs are still in place, just in case someone is relying on those links in a way we aren't aware of.

As no new pages should be deployed on graphql.github.io, hopefully this branch can be considered archived.

Signed-off-by: Brian Warner <brian@bdwarner.com>


orta (approved):
Cool, yeah, this will help people get to the updated pages 👍🏻

PR #1028, opened 2021-03-29
Azkron .inner-content max-width:740px is too narrow and many code examples are not fully visible

Azkron:
…re not fully visible so I just removed it

Closes #<issue number>


Azkron:
Forgot to add ID to commit for the EasyCLA

PR #1029, opened 2021-03-29
.inner-content max-width740px; is too narrow and hides code examples so I removed it. Azkron

Azkron:


Urigo:
@Azkron can you see if that is still necessary on the new website?
If yes, can you please open a new PR?
thanks!

PR #1031, opened 2021-03-30
Fix specs URL.

lucasmichot:
## Description

Right now the 2 spec links point to `https://graphql.github.io/graphql-spec/` that redirects to `http://spec.graphql.org/` (insecure).

This PR fixes that and points directly to  `https://spec.graphql.org/` (secure).


lucasmichot:
@leebyron, for some reason the CLA link does not work for me.

Maybe you can fix the link yourself and close this PR?

PR #1034, opened 2021-04-04
docs: Updating Discord popular channels

Urigo:
@benjie @brianwarner a first try at this, not sure if it's the best but let's have the discussion over this PR

### DIFF
--- a/src/content/community/Community-Users.md
+++ b/src/content/community/Community-Users.md
@@ -39,7 +39,6 @@ Please be patient and polite. These are not explicitly user support channels, al
 #### Discord
 
 * [Everything GraphQL: Curated By The Guild](https://discord.gg/xud7bH9)
-* [Graphile](https://discord.gg/graphile)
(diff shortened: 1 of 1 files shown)


benjie:
I think Graphile should probably be moved under the Vendors section; I’m happy wit the other changes though.

Urigo:
@benjie do you think The Guild one should also be moved?
it suppose to be general GraphQL but it's ran by us and I guess we are a vendor?

Urigo:
ok I've moved Graphile to vendors.
I've left The Guild in general Discord because that's the actual purpose of the server, not things that are specific for The Guild (and also if one day we'll have an official GraphQL Discord we'll just move everything there).

I've placed Graphile above Prisma because Prisma is no longer GraphQL focused so I think it makes more sense for the user.

@brianwarner @benjie let me know if everything looks ok and I can merge.

benjie (approved):

brianwarner:
Looks good to me, and I'm glad to see we are continuing to refine these pages after taking them live. Thanks @Urigo!

PR #1045, opened 2021-04-15
Server GraphQL structure does not match the Query

RudiBoshoff:
Currently learning GraphQL and noticed an inconsistency in between:
```graphql
type Query {
  hero: Character
  droid(id: ID!): Droid
}
```
and 
```
query {
  hero {
    name
  }
  droid(id: "2000") {
    name
  }
}
```

Closes #<issue number>

### DIFF
--- a/src/content/learn/Learn-Schema.md
+++ b/src/content/learn/Learn-Schema.md
@@ -101,7 +101,7 @@ That means that the GraphQL service needs to have a `Query` type with `hero` and
 
 ```graphql
 type Query {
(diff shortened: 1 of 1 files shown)


benjie:
Hi @RudiBoshoff thanks for contributing! There’s actually no error here since the argument for the field `hero(episode: Episode): Character` is optional (it has no `!`) you can query it without arguments (which you must do without parentheses).

Urigo:
closing per @benjie 's comment

PR #1053, opened 2021-05-26
Fixed Validation example link

shan7030:
## Description

Fixed the link to starWarsValidation-test.js in Validation section which is recently migrated to .ts from .js


brianwarner:
Thank you! This fix is also included in #1056

PR #1072, opened 2021-07-26
Changed The Expected Output

Jac-Bo:
# Wrong Expected Code Output

## Description

> When following the tutorial for this page on the most recent version of graphql it outputted 
> > { data: [Object: null prototype] { hello: 'Hello world!' } }

>  rather than what the page says,
> > { data: { hello: 'Hello world!' } }

> So I updated it accordingly so that the expected output is,
> > { data: [Object: null prototype] { hello: 'Hello world!' } }

### DIFF
--- a/src/content/graphql-js/Tutorial-GettingStarted.md
+++ b/src/content/graphql-js/Tutorial-GettingStarted.md
@@ -54,7 +54,7 @@ node server.js
 You should see the GraphQL response printed out:
 
 ```javascript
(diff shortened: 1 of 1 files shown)


orta:
Thanks for the PR but I don't think we want this, the previous version shows the object structure which  I think is useful - the extra logging doesn't really add much.

Urigo:
thank you @Jac-Bo 
Closing per @orta's comment
If you think this is a wrong decision, let's continue discussing it

PR #1084, opened 2021-08-24
Brand Guidelines

leebyron:
Adds a page outlining brand and trademark guidelines, includes updated polished assets with corrected colors.

Updates assets in use and links into brand.


orta (commented):
Preview: https://deploy-preview-1084--graphql-org.netlify.app/brand

I love the page, it's great. Forcing the acknowledgement is an interesting call, wonder what that looks like in the long run. 

For TypeScript's branding page, I tried to make sure that the wording and descriptions would hit search queries like "typescript logo" etc, might be worth a think if there's ways to do the same here

Only thing I'm a bit eh on is that the link lives under the GraphQL foundation in the footer, it feels like that would be the branding guidelines for the _foundation_ and not necessarily for GraphQL (and thus have the useful design assets folks are after.) 

Either way, very cool

orta (approved):

src/pages/brand.tsx
```diff
@@ -0,0 +1,446 @@
...
+              the&nbsp;GraphQL&nbsp;wordmark.
+            </li>
+          </ul>
+        </div>
+      </section>
+      <section>
+        <h2>Logo &amp; Hexagraph</h2>
+        <p>
+          Our logomark is called the Hexagraph. The GraphQL logo is composed of
+          the hexagraph and the GraphQL wordmark set in Rubik Light.
```
benjie: Capitalisation consistency: `Hexagraph` or `hexagraph`? I went lower case as it's commonly used in lower case further down the page, and used quotes to make the definition clearer instead of the capital.

```suggestion
          Our logomark is called the “hexagraph”. The GraphQL logo is composed of
          the hexagraph and the GraphQL wordmark set in Rubik Light.
```

src/pages/brand.tsx
```diff
@@ -0,0 +1,446 @@
...
+        {...(!agree
+          ? { className: "disabled" }
+          : { href: "/img/brand/graphql-brand-assets.zip" })}
+      >
+        Download Assets
+      </a>
+      <a
+        {...(!agree
+          ? { className: "disabled" }
+          : { href: "mailto:info@graphql.org?subject=Trademark+Request" })}
```
benjie: Support for `+` in `mailto:` links is unreliable and will often carry the `+` through into the subject of the relevant mail client - this happens for the default iOS mail client for example. Use percent encoding instead:

```suggestion
          : { href: "mailto:info@graphql.org?subject=Trademark%20Request" })}
```

benjie (approved):
LGTM, couple minor notes :+1:

src/pages/brand.tsx
leebyron: TIL! This appears in a few spots so let me resolve them all

leebyron:
Great feedback, thanks both of you!

Such a good point about SEO and key terms. I did a bit of rephrasing to hopefully do a better job of that. I also just pulled out a separate PR I'll rebase over in #1089 which gives this a clearer place in the footer.

orta (approved):
Yeah, solid improvement 👍🏻

PR #1089, opened 2021-08-25
Update footer links

leebyron:
Includes some of the newer additions to the site and minor rearranging and renaming of links to make more sense.


leebyron:
Before:
Screen Shot 2021-08-26 at 10 42 09 AM

After:
Screen Shot 2021-08-26 at 11 09 20 AM

PR #1102, opened 2021-09-01
Stricter GraphQL typing using gatsby-plugin-typegen

cometkim:
## Description

Hi :wave:

Glad to see that the GraphQL homepage is built with Gatsby.

I set up the [gatsby-plugin-typegen](https://github.com/cometkim/gatsby-plugin-typegen), which is the integration of GCG and Gatsby

All named queries are typed while gatsby develop running

## Further steps

To unlock the power of GraphQL, I explored best practices while developing and using this plugin, and here are my two cents.

Nullability for implicitly inferred fields is normal. To get rid of it, we can define our model via schema customization API in gatsby-node and inject pre-transformed data.

The build-time data layer is a blessing for both development and runtime.

Colocating data dependencies using fragments is even better. just like Relay does

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ yarn-error.log
 
 # Swap files
 *.swp
+
+# Codegen stuff
+src/__generated__/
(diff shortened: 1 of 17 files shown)


Urigo:
Thank you @cometkim ! what a great tool
looking forward for your next PRs!

PR #1125, opened 2021-09-22
refactor blog posts

cometkim:
## Description

@Urigo As I mentioned in https://github.com/graphql/graphql.github.io/pull/1102

We can safely remove most of the runtime validation by using the [Gatsby's Data Layer](https://www.gatsbyjs.com/docs/reference/graphql-data-layer/).

And this is an example of how to do it, by adding a [custom schema](https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/) for type `BlogPost` .

Reducing runtime code and using Gatsby Link correctly is actually significantly speeds up the navigation.


Urigo:
Wow @cometkim that is nice!
And also noticeably faster!
Thank you!

PR #1155, opened 2021-11-12
Fix incorrect query on Queries and Mutation

### LINKED ISSUE #1126: `Buffer is not defined` error in Variables example
### Description

Seeing the error `Buffer is not defined` in this query: https://graphql.org/learn/queries/#using-variables-inside-fragments

### Steps to Reproduce

Visit https://graphql.org/learn/queries/#using-variables-inside-fragments

### Expected Result 

The query would work as expected
image

dschafer:
Closes #1126

## Description

One of the queries here uses "friendsConnection", but the rest use "friends", and only friends is valid in this schema.

This replaces the incorrect usage.

Note that I'm not sure how to test the Graphiql portion of this change — the only validation I did was that the query looks correct now.

### DIFF
--- a/src/content/learn/Learn-Queries.md
+++ b/src/content/learn/Learn-Queries.md
@@ -130,7 +130,7 @@ It is possible for fragments to access variables declared in the query or mutati
 
 ```graphql
 # { "graphiql": true }
-query HeroComparison($first: Int = 3) {
+query HeroComparison {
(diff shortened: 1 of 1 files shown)


bretonium:
Although it fixes the error, it seem to completely remove variables from the fragment, and the section is about variables in the fragments.

dschafer:
Good point! I'll leave to someone to debug why the query itself is failing then.

PR #1175, opened 2021-12-04
Renamed to Dice

siyavashhamdi:


leebyron:
Dice being plural seems to make this change less clear than before.

orta:
I agree with Lee, thanks for giving it a shot @siyavashhamdi but I'm going to close this PR

PR #1176, opened 2021-12-10
Update graphql-ruby example code

rmosolgo:
GraphQL-Ruby has had some schema definition updates, so I thought I'd update this doc.


orta:
Looks reasonable to me, thanks

PR #1179, opened 2021-12-15
fix: use correct links to graphql spec/foundation

notrab:
## Description

Update links to correct headings on page for specification & foundation.


orta:
Thanks

PR #1186, opened 2022-01-08
update to use object with graphql() vs 3 params

thall1961:
Closes #<issue number>

## Description

Small update to the NodeJS GraphQL docs to use an object with the graphql() function rather than three parameters.


src/content/graphql-js/Tutorial-GettingStarted.md
```diff
@@ -33,14 +33,18 @@ var schema = buildSchema(`
 `);
 
 // The root provides a resolver function for each API endpoint
```
MichaelBailly: ```// the rootValue```

orta:
Thanks!

PR #1203, opened 2022-01-25
Add qlient as a python client.

qlient-org:
## Description

Add a new graphql python client.


orta:
Thanks, looks good to me 👍🏻

PR #1207, opened 2022-02-16
rename input type identifier in example on Learn-Schema page

tbcs:
## Description

Rename input type `ReviewInput` to `Review` to align the identifier with
the style used elsewhere (no Hungarian notation).


dimaMachina:
@Urigo can be closed, adding input suffix is a good practice

PR #1239, opened 2022-05-21
Fixed javascript code snippet

goncharov-vlad:
## Description

The snippet was not correct, it throws this error:
Error: Expected undefined to be a GraphQL schema.

The reason is that we passed configuration as params of qraphql function, but the function takes only one param - an object of the parameters.

See type declaration of the function by permalink bellow
https://github.com/graphql/graphql-js/blob/c7d7026982ceee536900a24ae31235127560297a/src/graphql.ts#L70


orta:
This looks good to me, thanks for the writeup

PR #1240, opened 2022-05-23
fix: fix marked package import

2239559319:
fix `marked` package import

## Description
In `Introspection` section, entering `queryType` in the input box will throw an error as shown in the figure. 
image
The reason is that the marked package used does not have a default export


orta:
Probably came through on an auto-update PR, thanks!

PR #1242, opened 2022-06-03
Fix the deprecation link

martinbonnin:


orta:
Confirmed that works - thanks

PR #1248, opened 2022-06-13
added styling for sticky sidebar

### LINKED ISSUE #1247: Sticky Sidebar for better reading
### Description

When a reader scrolls to the bottom of the doc, they have to go all the way to the top to navigate to other doc links. If the sidebar is sticky, it will be a good UX. 

### Motivation

When I was reading the doc, in order to see the options available in the sidebar, once I finished a particular page, I would have to scroll all the way to top again. This becomes annoying if we keep on repeating, so I used to edit the style property from the dev tools.

### Collaboration

I want to build it out. 

### Additional Context

This will provide a good UX.

Lokprakash-babu:
Closes #1247 

## Description

Added a sticky container and corresponding stylings to keep the sidebar sticked.

### DIFF
--- a/src/assets/css/_css/docs.less
+++ b/src/assets/css/_css/docs.less
@@ -249,12 +249,15 @@
   margin: 1em 0 4em;
   width: 195px;
   padding-left: 20px;
-
   @media screen and (max-width: 740px) {
     width: auto;
     padding-left: 0;
   }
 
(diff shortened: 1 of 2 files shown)


orta:
Thanks for the PR, but I'm afraid it's too long to have as a sticky element on the screen

PR #1250, opened 2022-06-16
Add genqlient to list of Go clients

### LINKED ISSUE #201: Add to graphql.org/code
There's a list of Go clients at https://graphql.org/code/#go-client, we should add genqlient! I don't know the process but I assume it's doable and might be a good way for people to find genqlient.

benjaminjkraft:
genqlient has been public for almost a year now, so it seems
like time to add it to this page! I mostly copied the descriptions
from its README.

Fixes Khan/genqlient#201.


orta:
👍🏻

PR #1267, opened 2022-09-09
Add Escape.tech – GraphQL Security, GraphQL.Security & GraphQL Armor

iCarossio:
Add Escape.tech – GraphQL Security, GraphQL.Security & GraphQL Armor :)

PR #1279, opened 2022-11-10
graphql http blog post

leebyron:


hwillson:
👋 @leebyron - I just happened to notice this fail. Looks like this was caused by https://github.com/graphql/graphql.github.io/pull/1267. The Netlify logs are reporting:

CleanShot 2022-11-09 at 19 57 08

Looks like we just need to fix the `github` value. I can submit a PR if that's helpful, and sorry I didn't notice this earlier!

hwillson:
Haha disregard - I just saw https://github.com/graphql/graphql.github.io/pull/1280. Thanks!

PR #1285, opened 2022-11-11
feat: add wundergraph to gateway section

jensneuse:
## Description

As discussed with Uri, this adds WunderGraph to the newly created Gateway section.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -183,6 +183,7 @@ exports.onCreatePage = async ({ page, actions }) => {
       otherLibraries: {
         Services: codeData.Services,
         Tools: sortedTools,
+        Gateways: codeData.Gateways,
         "More Stuff": codeData["More Stuff"],
       },
       languageList: languageList.sort((a, b) => {
--- a/src/content/code/gateways/wundergraph.md
+++ b/src/content/code/gateways/wundergraph.md
@@ -1,0 +1,86 @@
+---
+name: WunderGraph
+description: WunderGraph is an open-source GraphQL Gateway that is able to compose Apollo Federation, GraphQL, REST APIs, Databases, Kafka and more.
+url: https://wundergraph.com
+github: wundergraph/wundergraph
+---
+[WunderGraph](https://wundergraph.com) composes all your APIs into a single unified GraphQL API and 
+allows you to expose your Graph as a [secure and type-safe JSON-RPC API](https://docs.wundergraph.com/docs/features/graphql-to-json-rpc-compiler).
+
+To get started with WunderGraph, you can use `create-wundergraph-app` to bootstrap a new project:
+
+```bash
+npx create-wundergraph-app my-project -E nextjs-swr
+```
+
+On the client side, WunderGraph's JSON-RPC API integrates very well with frameworks like [Next.js, SWR](https://github.com/wundergraph/wundergraph/tree/main/examples/nextjs-swr) and React Query,
+while one the backend, we're able to leverage the power of "Server-Side-Only GraphQL".
+Handle authentication, authorization, validation, joins and more right in the Query Layer.
+
+```graphql
+mutation (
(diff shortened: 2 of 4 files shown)


hwillson:
👋 @jensneuse - thanks for this!

Hmmm - I wonder if creating a new "Gateways" section on this page is getting a bit too specific, and breaks the more high-level type of bucketing in place on the page. Gateways are a pretty specific subset of "Tools", and there aren't (currently) a lot of GraphQL gateways out there, so I'm not sure if calling them out separately will add much value to folks visiting the page. Don't get me wrong, I will happily submit a PR to add Apollo Router if we want to create this new category 😂, but I think this extra category might be splitting things up a bit too much. That being said, I'd love to hear what others in @graphql/content think.

Urigo:
Great ideas @jensneuse and @hwillson !

Adding a bit more context on the last overall we've mode for the code section, we had thought initially to create a couple of sub-categories under tools, but at the time there weren't enough, so we thought to leave it as general "tools" section and see when new things will up to differentiate.

I think Gateways/Supergraphs looks like an important category and the fact that Apollo Router, Schema Stitching and other very popular and important tools are not represented here points to the fact we might want to add that section.

I guess the open question for me would be if we want to add it to the top category (next to `tools`) or a subcategory of `tools`.
Tools currently is focused on things that are language agnostic, which I believe Gateways and Supergraphs are, so if we follow that direction, it means as a subcategory under tools.

Also I can think of a couple of new subcategories of language agnostic tools, so I guess my vote is to put it as a first subsection under `Tools` and place all existing links under `general` subsection.

Also, personally I would name this category `Gateways and Supergraphs` as any GraphQL library could act as a gateway and I think the distinguishing factor here is a tool that orchestrates underlying, remote GraphQL servers, so the rename would make it more clear.

I will follow up on this PR with other tools that fit that subcategory and at least one more sub category.

@jensneuse @hwillson does that fit your directions?

jensneuse:
> Great ideas @jensneuse and @hwillson !
> 
> Adding a bit more context on the last overall we've mode for the code section, we had thought initially to create a couple of sub-categories under tools, but at the time there weren't enough, so we thought to leave it as general "tools" section and see when new things will up to differentiate.
> 
> I think Gateways/Supergraphs looks like an important category and the fact that Apollo Router, Schema Stitching and other very popular and important tools are not represented here points to the fact we might want to add that section.
> 
> I guess the open question for me would be if we want to add it to the top category (next to `tools`) or a subcategory of `tools`.
> Tools currently is focused on things that are language agnostic, which I believe Gateways and Supergraphs are, so if we follow that direction, it means as a subcategory under tools.
> 
> Also I can think of a couple of new subcategories of language agnostic tools, so I guess my vote is to put it as a first subsection under `Tools` and place all existing links under `general` subsection.
> 
> Also, personally I would name this category `Gateways and Supergraphs` as any GraphQL library could act as a gateway and I think the distinguishing factor here is a tool that orchestrates underlying, remote GraphQL servers, so the rename would make it more clear.
> 
> I will follow up on this PR with other tools that fit that subcategory and at least one more sub category.
> 
> @jensneuse @hwillson does that fit your directions?

That's in line with my thinking. What should we do with this PR? Should I make changes as you proposed? Do you want to take over and arrange the content properly? I have to admit I don't feel 100% comfortable with the codebase and don't want to introduce a new style or patterns when introducing this new category.

Urigo:
sure, if you are not comfortable I can send a new PR instead.
Let's wait for @hwillson's opinion first

hwillson:
Thanks for the additional background @Urigo - this all sounds great to me!

jensneuse:
> sure, if you are not comfortable I can send a new PR instead. Let's wait for @hwillson's opinion first

Ok, then I'd welcome if you can take over incorporating this PR into the new structure. Thanks.

Urigo:
done in https://github.com/graphql/graphql.github.io/pull/1330

PR #1315, opened 2022-11-22
GraphQL Discord is now well established

benjie:
## Description

The GraphQL Discord was adopted September last year; it was previously the `Everything GraphQL: Curated By The Guild` server (i.e. they're the same place). So I've removed the redundant link, and I've also removed the link to the old Slack because people should not be going there any more (and we don't have any moderation of it).


hwillson (approved):
Thanks @benjie!

PR #1347, opened 2023-01-09
Remove `Express GraphQL`

TuvalSimha:
Signed-off-by: Tuval Simha <tuval.simha@gmail.com>

Closes #<issue number>

## Description

Remove Express GraphQL md file.
@Urigo


Urigo:
now that express-graphql [has been deprecated](https://github.com/graphql/express-graphql#this-library-is-deprecated), it might be good to remove it from the recommended frameworks
I've added some reviewers for confirmations
cc @leebyron @hwillson @benjie @michaelstaib

benjie:
Rather than deleting it (breaking permalinks, etc) can I suggest that we just change it to deprecated and remove copy saying it's the reference? Maybe even point to graphql-http from it?

Urigo:
@benjie in this case, it will just remove express-graphql from the list of tools, not from tutorials and guides, so I don't think it breaks any permalinks?

benjie:
Ah; I had assumed it was linkable content. If it's not then sure 👍

leebyron:
Works for me. @Urigo I trust you to merge changes like this if you think they are right 👍

PR #1353, opened 2023-01-14
Announcing contributed custom scalar specs at scalars.graphql.org

dondonz:
## Description

Announcement blog for contributed custom scalar specs at scalars.graphql.org


leebyron:
Nice work! Woohoo!

dondonz:
Thanks for all your work @andimarek and @leebyron!!!

andimarek:
@leebyron can we send out a tweet about this? thanks

PR #1379, opened 2023-03-19
Remove Slack

benjie:
There's no historical data in Slack due to it's lack of data retention and the Discord move was ages ago.


Urigo:
thanks @benjie

PR #1395, opened 2023-04-07
Add Correct Preview Agreement to foundation/join

jorydotcom:
## Description

I noticed that the participation agreement we link to as a preview on the foundation site is the old copy. PR adds correct preview agreement.


Urigo:
thanks @jorydotcom !

PR #1419, opened 2023-04-18
fix conf page footer alignment

beerose:
The footer alignment on the conf page was a bit off. (Mobile view stays the same)

Before:
CleanShot 2023-04-18 at 11 10 47@2x

After:
CleanShot 2023-04-18 at 11 19 57@2x


Urigo:
thank you @beerose !!

PR #1422, opened 2023-04-21
sorting languages

### LINKED ISSUE #1107: add toggle for alphabetical sort of language list in /code
### Description

only one html/css change requred, reference implementation:
https://codepen.io/dym-sh/pen/gORLrGQ?editors=1100

- also add Dart as alias for Flutter
- also set column-count to auto, no need for screen-size-breakpoints dependency

### Motivation

so, in https://graphql.org/code/
i was scanning the list for Dart, but only found Flutter after i started going one-by-one
both of which would be way faster if list would be in alphabetical order and had Dart

### Collaboration

sure.. not sure if its something important enough, so just an issue atm, i can prob send PR within a week

Lokprakash-babu:
Closes <https://github.com/graphql/graphql.github.io/issues/1107>

## Description
- Added sorting functionality based on popularity and alphabetical order.

### DIFF
--- a/src/assets/css/_css/code.less
+++ b/src/assets/css/_css/code.less
@@ -94,6 +94,27 @@
       }
     }
   }
+  .sortByContainer{
+    display: flex;
+    flex-flow: row wrap;
+    gap: 1rem;
+    align-items: center;
+    .sortBy{
+      font-weight: 500;
+    }
+    .inputContainer{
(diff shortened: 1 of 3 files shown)


TuvalSimha (approved):

Urigo:
thank you @Lokprakash-babu !

PR #1433, opened 2023-05-04
Add python to list of alternative languages

patrick91:
I hope this is ok 😊


Urigo:
thank you @patrick91 !

PR #1481, opened 2023-06-21
Add Program page and remove references to the speak page

beerose:


benjie:
Will this actually delete the speak page? I think it would be better to keep the speak page (to avoid breaking links) and update it to indicate that the CFP is now closed.

beerose:
> Will this actually delete the speak page? I think it would be better to keep the speak page (to avoid breaking links) and update it to indicate that the CFP is now closed.

Makes sense. I brought this page back

PR #1502, opened 2023-07-21
Restore /swapi-graphql endpoint via reverse proxy

acao:
Closes #1427

## Description

restore 200 reverse proxy rewrite logic for swapi-graphql

PR #1503, opened 2023-07-21
restore  legacy 301 redirects

acao:
## Description
follow up to #1502 - finish porting netlify `_redirects` -> gatsby cloud `createRedirect()`


Urigo:
thank you @acao !

PR #1551, opened 2023-09-11
Add Gato GraphQL to PHP servers

leoloso:
## Description

This PR adds [Gato GraphQL](https://github.com/GatoGraphQL/GatoGraphQL) to the PHP servers.

It also removes the previous 2 versions of the server (repo `leoloso/PoP` is now `GatoGraphQL/GatoGraphQL`, so they all point to the same source):

- GraphQL API for WordPress (which has been renamed to Gato GraphQL)
- GraphQL by PoP (deprecated)


Urigo:
thank you @leoloso

PR #1557, opened 2023-09-13
Update quicktype.md

pieh:
Seems like https://github.com/quicktype/quicktype is now https://github.com/glideapps/quicktype . Using old org name seems to result in GitHub API giving back empty result which seems to crash the builds

Closes #<issue number>


pieh:
Closing in favor of https://github.com/graphql/graphql.github.io/pull/1558

PR #1583, opened 2023-10-05
fix canonical

dimaMachina:


src/components/Conf/Seo/index.tsx
```diff
@@ -41,6 +42,9 @@ export default function SeoConf(props: {
       <meta name="twitter:title" content={twitterTitle} />
       <meta name="twitter:description" content={description} />
       <meta name="twitter:image" content={image} />
+      {props.pathname && (
+        <link rel="canonical" href={`https://graphql.org/${props.pathname}`} />
```
benjie: If `pathname` starts with a `/` then this will be an unpleasant URL. 
```suggestion
        <link rel="canonical" href={`https://graphql.org/${props.pathname.replace(/^[/]+/, '')}`} />
```

dimaMachina: previous solution didn't work, I replaced by gatsby plugin 
image

PR #1586, opened 2023-10-13
Update fragment's definition in Learn-Queries.md

inlineblock:
Updating the wording on how fragment should be used to nudge people towards a more scalable approach of using fragments to define data requirements.

## Description

Even though the spec allows fragments to be used however people see fit, the document should nudge people towards a better usage pattern that scales well. In this case, attempt to describe them as composable pieces of data requirements used by UI or utility code. I hope this isn't controversial.

### DIFF
--- a/src/content/learn/Learn-Queries.md
+++ b/src/content/learn/Learn-Queries.md
@@ -110,6 +110,7 @@ That's why GraphQL includes reusable units called _fragments_. Fragments let you
   }
 }
 
+# used by the comparison UI component
 fragment comparisonFields on Character {
   name
   appearsIn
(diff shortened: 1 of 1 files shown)


src/content/learn/Learn-Queries.md
```diff
@@ -110,6 +110,7 @@ That's why GraphQL includes reusable units called _fragments_. Fragments let you
   }
 }
 
+# used by the comparison UI component
```
benjie:
```suggestion
# used by the character-rendering UI component
```

benjie (changes requested):
Looks like a great improvement to me! My one comment is the text describes a comparison page rendering two characters side by side, so it’s the “character” component that would have the fragment I think. Maybe renaming the fragment would make sense but I’ve not looked at the rest of the document; for now I suggest a tiny tweak to your edits.

inlineblock:
@benjie thank you for the review! I can do a follow PR to rename the fragment across the document.

benjie (approved):

saihaj:
Hey @inlineblock  can you please rebase this PR?

PR #1619, opened 2024-02-05
fix: vercel rewrites

YassinEldeeb:


saihaj:
we can re-open this at a later to optimize it cc @dimaMachina

PR #1635, opened 2024-03-28
remove old `source` files

dimaMachina:
The new one lives in `website/src` (will be moved to `/src` after)


YassinEldeeb (approved):

YassinEldeeb:
prettier needs to be fixed

PR #1673, opened 2024-03-30
move `Spec` link to community

dimaMachina:
image


Urigo:
@graphql/tsc what do you think?
We want to clear the header and move some things to more internal resources,
Would moving the spec from the menu be ok?

benjie:
I think the Spec link should have higher priority than FAQ; perhaps move FAQ to community instead?

PR #1697, opened 2024-05-09
docs: Add more links from queries to schema

bubaflub:
## Description

Adds two more links from the "Queries and Mutations" learning page to the "Schemas and Types" pages where I think it would be relevant for folks to want to be able to click through and learn more.


Urigo:
Thank you @bubaflub !

PR #1703, opened 2024-05-17
Argument need not be optional to have a default value defined

gpj1:
Closes #<issue number>

## Description
Current documentation about Arguments says that _"When an argument is optional, we can define a default value ..."_
This seems to be misleading because a default value for an argument can be defined irrespective of optional or mandatory argument.

### DIFF
--- a/src/pages/learn/schema.mdx
+++ b/src/pages/learn/schema.mdx
@@ -63,7 +63,7 @@ type Starship {
 
 All arguments are named. Unlike languages like JavaScript and Python where functions take a list of ordered arguments, all arguments in GraphQL are passed by name specifically. In this case, the `length` field has one defined argument, `unit`.
 
-Arguments can be either required or optional. When an argument is optional, we can define a _default value_ - if the `unit` argument is not passed, it will be set to `METER` by default.
+Arguments can be either required or optional. We can define a _default value_ for an argument as well - if the `unit` argument is not passed, it will be set to `METER` by default.
 
 ## The Query and Mutation types
 


benjie:
Interesting nuance; technically you can’t define a default value for a non-optional argument, since an optional argument is one that is either nullable or has a default value, but this is a bit if a circular argument! Thanks for bringing this to our attention, definitely feels like the wording could be cleaned up 👍

src/pages/learn/schema.mdx
```diff
@@ -63,7 +63,7 @@ type Starship {
 
 All arguments are named. Unlike languages like JavaScript and Python where functions take a list of ordered arguments, all arguments in GraphQL are passed by name specifically. In this case, the `length` field has one defined argument, `unit`.
 
-Arguments can be either required or optional. When an argument is optional, we can define a _default value_ - if the `unit` argument is not passed, it will be set to `METER` by default.
+Arguments can be either required or optional. We can define a _default value_ for an argument as well - if the `unit` argument is not passed, it will be set to `METER` by default.
```
benjie, line 66: How about:

```suggestion
Arguments can be either required or optional. An argument is optional if its type is nullable or if it defines a _default value_. For the `length` field above, if the `unit` argument is not passed, it will be set to `METER` by default.
```

I think the code above should also be adjusted (and this change reflected as appropriate in the rest of the page), though that is not essential:

```
  length(unit: LengthUnit! = METER): Float
```

benjie (changes requested):

src/pages/learn/schema.mdx
benjie: (The change being the addition of the `!`)

PR #1704, opened 2024-05-19
fix navbar links and combine sponsor/partner pages

YassinEldeeb:


Urigo:
@YassinEldeeb @jorydotcom I think the header is way over populated
image
Maybe we can find a better design solution for this?

YassinEldeeb:
ah, I didn’t think it’d wrap with smaller screens.

Is the register button really needed in the nav? I think the button in the hero section is a pretty good call to action.

YassinEldeeb:
oh, I think we can remove the “Register” link in the navbar, it’s redundant, it takes the user to cvent, which we don’t want.

YassinEldeeb:
ok, @Urigo how about now? it's still over populated in my opinion too, but that's one less link now

jorydotcom:
@YassinEldeeb The Conference Committee met this morning and we discussed the nav issue you & @Urigo highlighted. They have requested that:

- The GraphQLConf 2023 link move to the footer
- We drop the 'Speakers' item for the time being, since we don't have anything announced
- We make the 'Register' link an anchor tag to the "Pricing and Registration" section of the homepage

They also want to combine the Partner and Sponsor pages, but I think that's a bigger PR. The group was unanimous though that we should _not drop_ 'Register' from the top nav.

jorydotcom (changes requested):
See https://github.com/graphql/graphql.github.io/pull/1704#issuecomment-2120661483

YassinEldeeb:
Thanks for your feedback @jorydotcom! I’ll apply the proposed changes in this PR.

YassinEldeeb:
@Urigo @jorydotcom can you please review the current state of the PR? I have applied your suggestions.

Urigo:
Looks much better to me
@jorydotcom can you give it another review and also maybe ping the marketing team?

jorydotcom:
It looks so much better!! Thank you @dimaMachina !!!! <3

jorydotcom (approved):

Urigo:
@YassinEldeeb one last comment, if I'm not on the main page and I click "Register", it doesn't do anything

src/app/conf/2024/layout.tsx
```diff
@@ -41,22 +42,15 @@ export default function Layout({
...
         }
         links={[
-          { children: "FAQ", href: "/conf/2024/faq" },
-          { children: "Speak", href: "/conf/2024/speak" },
-          { children: "Register", href: "https://cvent.me/gk2dRw" },
-          { children: "Sponsor", href: "/conf/2024/sponsor" },
-          { children: "Partner", href: "/conf/2024/partner" },
+          { children: <span>FAQ</span>, href: "/conf/2024/faq" },
+          { children: <span>Speak</span>, href: "/conf/2024/speak" },
+          { children: <span>Register</span>, href: "#attend" },
```
YassinEldeeb:
```suggestion
          { children: <span>Register</span>, href: "/conf/2024#attend" },
```

YassinEldeeb:
@Urigo fixed it

Urigo:
thank you!

PR #1712, opened 2024-06-02
Update 2024 sponsors

YassinEldeeb:
we can wait on this if you want, so there are more, but this is the current list that I'm aware of according to the LF emails sent


src/app/conf/2024/sponsors.tsx
```diff
@@ -132,39 +96,59 @@ export function Sponsors() {
...
-          className="grid-cols-2 xl:grid-cols-4"
-          linkClassName="p-6 lg:p-10 h-28 lg:h-[155px]"
-        />
-        <h3 className={classes.heading}>Workshop Day Sponsor</h3>
-        <List
-          items={workshopDaySponsors}
-          className="grid-cols-2 lg:grid-cols-1 lg:w-1/2 mx-auto"
-          linkClassName="p-8 lg:p-10 h-28 lg:h-[155px]"
-        />
+        <h1 className={classes.title}>Thanks to our 2024 sponsors!</h1>
```
dimaMachina, line 87: Should it be just `Sponsors` ?

dimaMachina (approved):

src/app/conf/2024/sponsors.tsx
YassinEldeeb: @Urigo

PR #1725, opened 2024-07-10
GraphQL Local Initiative

benjie:


src/pages/community/foundation/local-initiative.mdx
```diff
@@ -0,0 +1,136 @@
...
+  - Sponsors
+  - Schedule
+  - Speaker/talk introduction
+- **Promotion** through GraphQL’s Twitter and LinkedIn channels.
+- **Custom artwork** adhering to the GraphQL Brand Guidelines; including
+  avatars, event banners, and slide templates. \*\*
+
+\* The content of swag bundles may change from time to time. At time of writing
+they are likely to consist of 2-3 t-shirts (sizes to be chosen by organizer),
+stickers, pin-badges, and a GraphQL bow-tie!
```
benjie: @jorydotcom what's the latest on this? I wrote it from memory... and I don't have the best memory! :grimacing:

src/pages/blog/2024-07-22-graphql-local-initiative.mdx
```diff
@@ -0,0 +1,46 @@
...
+---
+
+The GraphQL Foundation's primary mission is to ensure the continued evolution of
+the GraphQL specification and reference implementations. As part of this
+mission, we strive to provide collaboration infrastructure that fosters a
+culture of community and innovation. Local get-togethers play a crucial role in
+this, bringing together developers, library maintainers, vendors, and
+enthusiasts to share knowledge and collaborate.
+
+This is why we're pleased to announce the launch of the **GraphQL Local
```
jorydotcom: I wonder if we want to say something like, "That's why we've expanded our support for local communities with the launch of the GraphQL Local Initiative" - so we're acknowledging the past efforts of the meetup pro network but also saying, this replaces that program

src/pages/community/foundation/local-initiative.mdx
```diff
@@ -0,0 +1,136 @@
...
+  **[GraphQL Brand Guidelines](https://graphql.org/brand/)**.
+- The Local accepts **GraphQL TSC oversight**:
+  - At least one
+    [GraphQL Technical Steering Committee (TSC) member](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#tsc-members)
+    must have administrator access to the Local’s online platforms.
+  - In case of a serious Code of Conduct breach, an organizer may be replaced.
+- The Local has 1 or more organizer from the area. (We're pretty flexible on the
+  definition of "from the area", just be reasonable.)
+- Each official event will have at least least one **organizer present**.
+- Official events will be held in an appropriate **professional setting**.
```
jorydotcom, line 46: Should we clarify with some examples? E.g. would a pub qualify, or a games arcade?

src/pages/community/foundation/local-initiative.mdx
jorydotcom: We will give the meetup organizer a code, and right now the bundle is set up to be 3 t-shirts, 100 stickers, 10 enamel pins and 1 bow tie! I like keeping it a bit more generic now, though, as inventory may change

src/pages/community/foundation/local-initiative.mdx
```diff
@@ -0,0 +1,136 @@
...
+- **Swag bundle**\* to distribute to speakers, attendees, or organizers (max 1
+  per quarter).
+- **Google Slides templates** for:
+  - Code of Conduct
+  - Call for Papers (CFP)
+  - Sponsors
+  - Schedule
+  - Speaker/talk introduction
+- **Promotion** through GraphQL’s Twitter and LinkedIn channels.
+- **Custom artwork** adhering to the GraphQL Brand Guidelines; including
```
jorydotcom: I think we want to be clear that this is not, "we will make custom graphics for each of your meetups" - I think we should just call this Local Artwork so that we don't indicate that we are going to do custom graphic design work

src/pages/community/foundation/local-initiative.mdx
```diff
@@ -0,0 +1,136 @@
...
+
+Organizers of a Local must agree:
+
+- All official events and related channels will be governed by the
+  **[GraphQL Code of Conduct](https://graphql.org/codeofconduct/)**.
+- Official communications will comply with the
+  **[GraphQL Brand Guidelines](https://graphql.org/brand/)**.
+- The Local accepts **GraphQL TSC oversight**:
+  - At least one
+    [GraphQL Technical Steering Committee (TSC) member](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#tsc-members)
```
jorydotcom, line 38: I think we need this to just be you; I am a little concerned this might feel like a big hill if the Local person doesn't know a TSC member and may be afraid to ask someone they don't know to be an admin

src/pages/community/foundation/local-initiative.mdx
```diff
@@ -0,0 +1,136 @@
...
+## Additional Support
+
+If you need extra support for your event (e.g., transportation costs or higher
+venue fees/larger food and drinks budget for larger get-togethers), individual
+GraphQL Foundation member organizations may be able to offer additional support,
+often with a small ask such as saying a few words about them during the event.
+If your gathering is approved, you will be provided with contact information for
+member organizations who have expressed interest in supporting gatherings.
+
+## Apply now!
```
jorydotcom: Do we want to say "Apply" or "Sign Up"? We had been talking like this is a "sign up and you get it, but you have to adhere to these to stay on" kind of thing.

jorydotcom (commented):
I have a few comments / thoughts to refine this a little!

src/pages/community/foundation/local-initiative.mdx
benjie: I'm happy to leave this open, if they have questions/venue in mind then they can ask. For example, I think the main room of a pub is probably not an appropriate setting, but exclusive use of the events room in a pub is probably fine.

src/pages/community/foundation/local-initiative.mdx
benjie: Good point; have renamed to "Local artwork" throughout.

src/pages/community/foundation/local-initiative.mdx
benjie: That's a good point, I hadn't considered that they might think they need to ask a TSC member directly. I've added parenthesis to indicate that _we_ will assign a TSC member to make this clearer. I was going to add that they could request a specific TSC member but figured that this is unnecessary - also you're right it'll probably be me for most of them.

src/pages/community/foundation/local-initiative.mdx
benjie: I've modified the language, I don't think "sign up" is quite right either since there is a process albeit a short one. Let me know what you think of the alternative wording.

benjie:
Thanks for the input @jorydotcom! I think I've addressed your feedback, and also added another couple minor changes - let me know what you think? (It's all in that last commit.) In particular based on feedback by @jemgillam I wanted to make it explicit that we support both new and existing meets.

bignimbus:
LGTM!

jorydotcom (approved):

PR #1731, opened 2024-07-26
"Exploring true nullability" blog post

benjie:
This is just a draft and needs discussion with the nullability WG before being posted.


martinbonnin:
@benjie should we close this one?

benjie:
I just scanned over it and I think it's reasonably close to where we went in the end. I'll add it to my (long) todo list, and see if we can get it updated and posted. Would be great to pair it with the release of `onError` in GraphQL.js.

PR #1775, opened 2024-09-12
Create Semantic Nullability Blog Post

twof:


leebyron:
Love it! Ship whenever you feel this is ready

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
+---
+title: "Semantic Nullability For Application Developers"
+tags: ["spec"]
+date: 2024-09-12
+byline: Alex Reilly
+---
+
+# Semantic Nullability
+
+> This blog post is directed at application developers using GraphQL. If you are a library author, you should read the more detailed feature spec instead.
```
benjie, line 10: This should either be a link to said spec or a promise to deliver such (e.g. via follow up blog post)

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+```graphql
+@extendedNullability
+
+type User { 
+  id: ID!
+  name: String!
+  age: Int
+  posts: [Post]
+}
```
benjie, line 70: This example (and the previous one) should include a nullable field too, perhaps avatarUrl?

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+```
+
+We can now trust that `age` and `posts` will never be null unless an error causes their values to fail to resolve.
+
+### Frontend migration
+Client libraries that take advantage of the new features in this release may provide a flags to alter their error handling behavior. In this example, `ApolloClient` is providing a configuration option that causes the access of a field with an error to throw.
```
benjie:
```suggestion
Client libraries that take advantage of the new features in this release may provide flags to alter their error handling behavior. In this hypothetical example, `ApolloClient` is providing a configuration option that causes the access of a field with an error to throw.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+```typescript
+const client = new ApolloClient({
+  ...
+  throwOnError: true
+});
+```
+
+Semantic Nullability gives clients and GraphQL tools more flexibility in how they respond to errors. Some may want to recreate the existing destructive behavior of GraphQL or do something specific to their domain like Relay's use of React error boundaries.
```
benjie:
```suggestion
Semantic Nullability gives clients and GraphQL tools more flexibility in how they respond to errors. Some may want to recreate the existing destructive behavior of GraphQL’s null bubbling, others may want to leverage their language’s native error handling mechanics, or even do something specific to their domain like Relay's use of React error boundaries.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+Because your client can decide how it handles errors, it will also be responsible for providing a modified version of the schema. For example, if the client raises an exception when an errored field is read, it can mark all "semantically non-null" fields as non-nullable in the schema provided to you as a frontend developer.
+
+![Example code generation](https://i.imgur.com/i3hdCND.png)
+
+You should use the modified version of the schema when doing code generation for your frontend application.
```
benjie:
```suggestion
The modified version of the schema can also incorporate any client-side schema extensions you may have. You should use the modified version of the schema when doing code generation for your frontend application.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+Altering the error handling behavior of your client may be a breaking change if your schema has already adopted semantic nullability, so it's suggested that you select new error handling behavior for your client first. Read the documentation for your specific client for more information.
```
benjie, line 93: This first sentence is a little confusing (changing client error handling may be a breaking change so you should change client error handling first); could you reword it (e.g. first -> before adopting semantic nullability) or expand it? Also changing client error handling is a breaking change whether or not you’ve adopted semantic nullability, I think? Also it’s somewhat independent of it: you can use one without the other? Perhaps something less specific might work well.

Also most clients don’t have this yet (do they?), should we be surfacing something more actionable?

benjie (changes requested):
Reads really well and a great introduction to the topic! I’ve included some suggested edits/typo fixes, and I wonder if we can make the final paragraph more actionable?

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+| String? | Nullable          |
+| String  | Semantic non-null |
+| String! | Strict non-null   |
+
+Types are now semantic non-null by default. Question marks are used to indicate a nullable field similar to many other modern languages. `String!` retains its meaning. This is of course, a breaking change, and GraphQL prides itself in offering a path to non-breaking evolution for existing services. So alongside the new type, we're introducing some mechanics to assist developers in making incremental updates to their applications.
+
+### Server migration
+Once Semantic Nullability has been released, you will be able to start migrating by updating your service to use the most recent version of GraphQL.
+
+This will open up the option to begin evolving your schema document by document. You can place the document directive `@SemanticNullability` at the top of a file to begin using the new nullability features in that file. The directive will not impact the interpretation of any other files in your schema.
```
BoD, line 58:
```suggestion
This will open up the option to begin evolving your schema document by document. You can place the document directive `@extendedNullability` at the top of a file to begin using the new nullability features in that file. The directive will not impact the interpretation of any other files in your schema.
```

Not 100% sure on which directive name we landed on, but `@extendedNullability` is used below

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+```graphql
+type User { 
+  id: ID! 
+  name: String!
+  age: Int
+  posts: [Post] 
+}
+```
+
+One thing to make clear: is this is not like using a type system in a compiled language (TS, Swift, Kotlin) where the type system makes compile-time guarantees about behavior at runtime. Rather you can think of GraphQL's type system as a "runtime" type system. You can trust that `age` will be an `Int`, because at runtime, GraphQL will assert that it is a `Int`, and throw an error if it is not. All types effectively represent a typecast, or a type assertion.
```
BoD:
```suggestion
One thing to make clear: this is not like using a type system in a compiled language (TS, Swift, Kotlin) where the type system makes compile-time guarantees about behavior at runtime. Rather you can think of GraphQL's type system as a "runtime" type system. You can trust that `age` will be an `Int`, because at runtime, GraphQL will assert that it is a `Int`, and throw an error if it is not. All types effectively represent a typecast, or a type assertion.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
benjie: Good catch! The capitalisation is off either way.

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
benjie: We don't really have a single place to point library authors to right now since we're "between implementations" - let's ask them to get involved?

```suggestion
> This blog post is directed at application developers using GraphQL. If you are a library author, the [nullability working group](https://github.com/graphql/nullability-wg) would love to hear from you - file an issue or join a meeting!
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+```graphql
+type User { 
+  id: ID! 
+  name: String!
+  age: Int
+  posts: [Post] 
+}
+```
+
+One thing to make clear: this is not like using a type system in a compiled language (TS, Swift, Kotlin) where the type system makes compile-time guarantees about behavior at runtime. Rather you can think of GraphQL's type system as a "runtime" type system. You can trust that `age` will be an `Int`, because at runtime, GraphQL will assert that it is a `Int`, and throw an error if it is not. All types effectively represent a typecast, or a type assertion.
```
benjie, line 25: Generally type casts are just trusted without verification; let's just go with "type assertion" alone.

```suggestion
One thing to make clear: this is not like using a type system in a compiled language (TS, Swift, Kotlin) where the type system makes compile-time guarantees about behavior at runtime. Rather you can think of GraphQL's type system as a "runtime" type system. You can trust that `age` will be an `Int`, because at runtime, GraphQL will assert that it is a `Int`, and throw an error if it is not. All types effectively represent a type assertion.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+This results in `null` having two different meanings in GraphQL.
+1. No value was provided. The `User` never provided an `age`.
+2. There was an error resolving the field. The `User` provided an age in the form of a `Float` and it couldn't be cast to an `Int`, or our new `BirthdayBoy` provider [microservice](https://www.youtube.com/watch?v=y8OnoxKotPQ) timed out and never returned the `User`'s birthday.
```
benjie, line 29: Slight restructure; Markdown wants a newline here and it felt safer to explicitly call out the second sentence in each list item as an example.

Feel free to reject this, but please put a blank line just above the numbered list.

```suggestion
This results in `null` having two different meanings in GraphQL:

1. no value was provided (e.g. the `User` never provided an `age`), or
2. there was an error resolving the field (e.g. the `User` provided an age in the form of a `Float` and it couldn't be cast to an `Int`, or our new `BirthdayBoy` provider [microservice](https://www.youtube.com/watch?v=y8OnoxKotPQ) timed out and never returned the `User`'s birthday).
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+We can differentiate between the two by calling the first `null`, and the second `(Error, null)`. The first type is signaled by a `null` return value and no associated error in the errors array. The second is signaled by a `null` return value and an associated error in the errors array.
+
+Returning to the GraphQL type system, we can see it has two options to indicate the nullability of a field.
+1. `String` which we now know means the field can be a `String`, `null`, or `(Error, null)`
+2. `String!` which we now know means the field can be `String`
```
benjie, line 35:
```suggestion
Returning to the GraphQL type system, we can see it has two options to indicate the nullability of a field:

1. `String`, which we now know means the field can be a `String`, `null`, or `(Error, null)`; and
2. `String!`, which we now know means the field can be `String`.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+When an error occurs resolving a `String` field, it's not much of a problem for clients. They can decide if they can deal with that field missing or not. However when an error occurs resolving a `String!` field, GraphQL responds by destroying part of the result data before it's sent to the client. Given the danger, many developers choose to never use any non-nullable fields.
```
benjie, line 37: I think you're meaning the server-side "don't make non-null fields available" idea, rather than the client-side "don't consume non-null fields".

```suggestion
When an error occurs resolving a `String` field, it's not much of a problem for clients. They can decide if they can deal with that field missing or not. However when an error occurs resolving a `String!` field, GraphQL responds by destroying part of the result data before it's sent to the client. Given the danger, many developers choose to never add non-nullable fields to the schema.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+Every time developers are surveyed about their issues with GraphQL, they talk about nullability. The Nullability Working Group has been hard at work, and we believe we finally have a solution.
```
benjie, line 39: We've had a number of solutions (including e.g. CCN); let's big up this one a little more:

```suggestion
Every time developers are surveyed about their issues with GraphQL, they talk about nullability. The Nullability Working Group has been hard at work, and we believe we finally have a workable solution that meets everyone's needs.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+The root of the problem is that developers want a way to express that all `User`s are expected to have an `age`, but if there is an error and GraphQL can't resolve `age`, then they'd like to deal with it client-side. In order to do that, they need the type system to allow for a third type of nullability.
+
+3. A field which can be `String` or `(Error, null)`
+
+This is what we're calling "Semantic non-null". The syntax we've chosen is the following
```
benjie, line 45:
```suggestion
This is what we're calling "Semantic non-null". The syntax we're currently proposing is the following
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+| Syntax  | Meaning           |
+| ------- | ----------------- |
+| String? | Nullable          |
+| String  | Semantic non-null |
+| String! | Strict non-null   |
+
+Types are now semantic non-null by default. Question marks are used to indicate a nullable field similar to many other modern languages. `String!` retains its meaning. This is of course, a breaking change, and GraphQL prides itself in offering a path to non-breaking evolution for existing services. So alongside the new type, we're introducing some mechanics to assist developers in making incremental updates to their applications.
```
benjie, line 53: Rather than saying this _is_ a breaking change, let's say it _would be_ a breaking change and then we can clarify why it isn't.

```suggestion
With this syntax, types become semantic non-null by default. Question marks (`?`) are used to indicate a nullable field similar to many other modern languages. `String!` retains its meaning.

This would be a breaking change; but GraphQL prides itself in offering a path to non-breaking evolution for existing services, so alongside the new type we're introducing some mechanics to assist developers in making incremental updates to their applications.
```

benjie (changes requested):
I've included some minor clarity/formatting tweaks; I still wonder if we should add a call to action: how do people get started?

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
glen-84, line 25: > You can trust that `age` will be an `Int`, because at runtime, GraphQL will assert that it is a `Int`, and throw an error if it is not.

`age` is nullable here, so it could be `null` as well, and won't throw for a `null` value. Maybe `name` would be a better example?

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
glen-84, line 37: > GraphQL responds by destroying part of the result data

`destroying` is quite a strong word. 🙂 Perhaps `nullifying`?

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
glen-84, line 25:
```suggestion
One thing to make clear: this is not like using a type system in a compiled language (TS, Swift, Kotlin) where the type system makes compile-time guarantees about behavior at runtime. Rather you can think of GraphQL's type system as a "runtime" type system. You can trust that `age` will be an `Int`, because at runtime, GraphQL will assert that it is an `Int`, and throw an error if it is not. All types effectively represent a typecast, or a type assertion.
```

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
glen-84, line 35: You switch here from talking about an Integer age field, to a String field? The same example should flow through.

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+This will open up the option to begin evolving your schema document by document. You can place the document directive `@SemanticNullability` at the top of a file to begin using the new nullability features in that file. The directive will not impact the interpretation of any other files in your schema.
+
+After migration, a `User` type would look like this.
+
+```graphql
+@extendedNullability
+
+type User { 
+  id: ID!
+  name: String!
```
glen-84, line 67: Maybe a note that the `!`'s can be removed, especially for non-ID fields?

src/pages/blog/2024-08-14-semantic-nullability-for-application-developers.mdx
```diff
@@ -0,0 +1,93 @@
...
+  ...
+  throwOnError: true
+});
+```
+
+Semantic Nullability gives clients and GraphQL tools more flexibility in how they respond to errors. Some may want to recreate the existing destructive behavior of GraphQL’s null bubbling, others may want to leverage their language’s native error handling mechanics, or even do something specific to their domain like Relay's use of React error boundaries.
+
+Because your client can decide how it handles errors, it will also be responsible for providing a modified version of the schema. For example, if the client raises an exception when an errored field is read, it can mark all "semantically non-null" fields as non-nullable in the schema provided to you as a frontend developer.
+
+![Example code generation](https://i.imgur.com/i3hdCND.png)
```
glen-84, line 89: Why is `dog` `String` and not `String?` on the RHS? It wouldn't become non-nullable because of the client library?

glen-84 (changes requested):

PR #1780, opened 2024-09-20
fix typo in pages/graphql-js/object-types.mdx

askariali:
## Description

This PR will fix a typo in the graphql-js/object-types page.
The typo: die -> dice

### DIFF
--- a/src/pages/graphql-js/object-types.mdx
+++ b/src/pages/graphql-js/object-types.mdx
@@ -12,7 +12,7 @@ type Query {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


benjie:
Thank you for the suggestion, but `dice` is a plural noun; the singular of `dice` is `die` (as in "the die was cast"), so `a dice` does not make grammatical sense.

https://www.merriam-webster.com/dictionary/die

askariali:
@benjie Oh, thanks for clarifying

PR #1787, opened 2024-10-02
Minor change to make fetch work

3willows:
Closes #<issue number>

### DIFF
--- a/src/pages/graphql-js/graphql-clients.mdx
+++ b/src/pages/graphql-js/graphql-clients.mdx
@@ -61,7 +61,7 @@ var query = /* GraphQL */`query RollDice($dice: Int!, $sides: Int) {
   rollDice(numDice: $dice, numSides: $sides)
 }`
 
-fetch("/graphql", {
+fetch("http://localhost:4000/graphql", {
   method: "POST",
   headers: {
     "Content-Type": "application/json",


benjie (changes requested):
Thank you for contributing! However, I'm not sure this is the right change:

- Assuming you're serving the code from http://localhost:4000 this should not be needed.
- If you're not serving the code from http://localhost:4000 then there are all sorts of additional concerns that we'd need to address, such as CORS.

Perhaps a clarifying note to state that this code must be served from your server at http://localhost:4000 would be a better solution?

3willows:
Hi!  So sorry for the late reply, this slipped my mind.

I also notice my issue was not very well written at all.

The PR concerns a page that is part of an ongoing tutorial.

At the immediate previous page, the text suggests the reader (as a matter of default) to start the server at port 4000.  

See bottom of this page: https://graphql.org/graphql-js/running-an-express-graphql-server/

Ditto the text on the page I want to change, immediately above,

> For example, let’s say we mounted a GraphQL server on http://localhost:4000/graphql as in the example code for [running an Express GraphQL server](https://graphql.org/graphql-js/running-an-express-graphql-server/), and we want to send the GraphQL query { hello }. We can do this from the command line with curl. If you paste this into a terminal: ...

As I was following along the tutorial (my [package.json](https://github.com/3willows/graphQL-quickStart/blob/main/package.json) is here, if it helps)

```
fetch("/graphql", {
```

returns this error.

```
node:internal/deps/undici/undici:12502
      Error.captureStackTrace(err, this);
            ^

TypeError: Failed to parse URL from /graphql
    at node:internal/deps/undici/undici:12502:13 {
  [cause]: TypeError: Invalid URL
      at new URL (node:internal/url:797:36)
      at new Request (node:internal/deps/undici/undici:4853:25)
      at fetch (node:internal/deps/undici/undici:9662:25)
      at fetch (node:internal/deps/undici/undici:12500:10)
      at value (node:internal/bootstrap/web/exposed-window-or-worker:72:12)
      at Object.<anonymous> (/Users/jasper/Desktop/graphQL-quickStart-main/client.js:15:1)
      at Module._compile (node:internal/modules/cjs/loader:1358:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
      at Module.load (node:internal/modules/cjs/loader:1208:32)
      at Module._load (node:internal/modules/cjs/loader:1024:12) {
    code: 'ERR_INVALID_URL',
    input: '/graphql'
  }
}

Node.js v20.15.0
```

Hence the suggested change.

benjie:
It seems you're running that from Node, but the text around that code explicitly says:

> It’s also simple to send GraphQL from the browser. Open up http://localhost:4000/graphql, open a developer console, and paste in: `fetch("/graphql", ...`

I.e. this command is only valid from in the browser.

3willows:
My fault for not reading properly, many apologies.

benjie:
No apology necessary! Do you think a comment in the code would make this clearer?

PR #1794, opened 2024-10-11
Make "Schemas and Types" page more approachable to newcomers

jonathanberger:
I'm new to GraphQL so I may be mistaken but I believe the author intended to capitalize the type names in this passage since they are capitalized in the paragraphs below. The capitalization is important to get right in this context since the example right above the changed text has both "query" and "Query" in it.

### DIFF
--- a/src/pages/learn/schema.mdx
+++ b/src/pages/learn/schema.mdx
@@ -67,7 +67,7 @@ Arguments can be either required or optional. When an argument is optional, we c
 
 ## The Query and Mutation types
 
-There are two special "entry point" types that by default are named `Query` and `Mutation`. These types are the same as a regular object type, but they are special because they define the _entry point_ of every GraphQL query. So if you see a query that looks like:
+Every GraphQL schema must support `query` operations, the _entry point_ for these operations is a regular object type that by default is called `Query`. So if you see a query that looks like:
 
 ```graphql
 # { "graphiql": true }
@@ -90,9 +90,18 @@ type Query {
 }
 ```
 
-Mutations work in a similar way - you define fields on the `Mutation` type, and those are available as the root mutation fields you can call in your query.
+Schemas may also choose to support `mutation` operations by adding another entry point type, by default called `Mutation`. Mutations work in a similar way to queries - you define fields on the `Mutation` type, and those are available as the root mutation fields you can call in your mutation operation.
 
 It's important to remember that other than the special status of being the entry points, the `Query` and `Mutation` types are the same as any other GraphQL object type, and their fields work exactly the same way.
+
+You can choose to name your entry point types differently; if you choose to do so then you will need to inform GraphQL of the new names using the `schema` keyword:
+
+```graphql
+schema {
+  query: MyQueryType
+  mutation: MyMutationType
+}
+```
 
 ## Scalar types
 
--- a/src/pages/learn/schema.mdx
+++ b/src/pages/learn/schema.mdx
@@ -76,7 +76,7 @@ schema {
 }
 ```
 
-Every GraphQL service has a `query` type and may or may not have a `mutation` type. These types are the same as a regular object type, but they are special because they define the _entry point_ of every GraphQL query. So if you see a query that looks like:
+Every GraphQL service has a `Query` type and may or may not have a `Mutation` type. These types are the same as a regular object type, but they are special because they define the _entry point_ of every GraphQL query. So if you see a query that looks like:
 
 ```graphql
 # { "graphiql": true }
--- a/src/pages/learn/schema.mdx
+++ b/src/pages/learn/schema.mdx
@@ -67,16 +67,7 @@ Arguments can be either required or optional. When an argument is optional, we c
 
 ## The Query and Mutation types
 
-Most types in your schema will just be normal object types, but there are two types that are special within a schema:
-
-```graphql
-schema {
-  query: Query
-  mutation: Mutation
-}
-```
-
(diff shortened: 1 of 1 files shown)


jorydotcom:
Thanks for contributing, @jonathanberger ! Can you sign our CLA?

benjie:
Hi @jonathanberger; thanks for your contribution! These terms can be overloaded quite heavily, making it confusing; I did some work on the spec to stop us using the word `query` in so many ways: https://github.com/graphql/graphql-spec/issues/715

I believe that the text here is not (contrary to what it says) referring to a `Query` type, but instead saying that there must be a type (it can have any valid name) that's used for the "query operation". We now call this the "root operation type" for the query operation: https://spec.graphql.org/draft/#sec-Root-Operation-Types

e.g. this is a valid schema:

```graphql
type Query {
  a: Int
}
```

This is also a valid schema (note it does not have a `Query` type, but it does have a type for the `query` operation, called `Root`):

```graphql
schema {
  query: Root
  mutation: Changes
}
type Root {
  a: Int
}
type Changes {
  b: Int
}
```

This is also valid (but super confusing!):

```graphql
schema {
  query: Mutation
  subscription: Query
}
"""The query root operation type in this schema"""
type Mutation {
  a: Int
}
"""The subscription root operation type in this schema"""
type Query {
  b: Int
}
```

I agree the text you've found is confusing; would you like to submit an improved fix that factors in this knowledge?

jonathanberger (commented):
Benjie, thanks so much. That really help clarifies. I took a crack at updating the docs to what I think I would have wanted before wading into this and understanding it a bit more. I realize I'm cutting out an example that is there today, but I think it's a beneficial simplification that will ease the GraphQL learning curve.

I had already been confused why this special word `schema` was being used without being fully explained anywhere else on the page, so I took out the example that used it.

src/pages/learn/schema.mdx
```diff
@@ -67,16 +67,7 @@ Arguments can be either required or optional. When an argument is optional, we c
...
-
-```graphql
-schema {
-  query: Query
-  mutation: Mutation
-}
-```
-
-Every GraphQL service has a `query` type and may or may not have a `mutation` type. These types are the same as a regular object type, but they are special because they define the _entry point_ of every GraphQL query. So if you see a query that looks like:
+There are two special "entry point" types that by default are named `Query` and `Mutation`. These types are the same as a regular object type, but they are special because they define the _entry point_ of every GraphQL query. So if you see a query that looks like:
```
benjie:
```suggestion
Every GraphQL schema must support `query` operations, the _entry point_ for these operations is a regular object type that by default is called `Query`. So if you see a query that looks like:
```

src/pages/learn/schema.mdx
```diff
@@ -101,7 +92,7 @@ type Query {
 
 Mutations work in a similar way - you define fields on the `Mutation` type, and those are available as the root mutation fields you can call in your query.
```
benjie:
```suggestion
Schemas may also choose to support `mutation` operations by adding another entry point type, by default called `Mutation`. Mutations work in a similar way to queries - you define fields on the `Mutation` type, and those are available as the root mutation fields you can call in your mutation operation.
```

benjie (changes requested):
This documentation is significantly out of date, we've not even mentioned subscriptions for example. I've taken a slightly alternative approach, what do you think?

src/pages/learn/schema.mdx
```diff
@@ -101,7 +92,7 @@ type Query {
...
 
-It's important to remember that other than the special status of being the "entry point" into the schema, the `Query` and `Mutation` types are the same as any other GraphQL object type, and their fields work exactly the same way.
+It's important to remember that other than the special status of being the entry points, the `Query` and `Mutation` types are the same as any other GraphQL object type, and their fields work exactly the same way.
```
benjie, line 95: > I had already been confused why this special word schema was being used without being fully explained anywhere else on the page, so I took out the example that used it.

I see what you mean; it's introduced a little to early. But since we talk about naming things _by default_ maybe it's best to reintroduce the `schema` keyword at the end of this section detailing what you need to do if you don't want the defaults:

````suggestion
It's important to remember that other than the special status of being the entry points, the `Query` and `Mutation` types are the same as any other GraphQL object type, and their fields work exactly the same way.

You can choose to name your entry point types differently; if you choose to do so then you will need to inform GraphQL of the new names using the `schema` keyword:

```graphql
schema {
  query: MyQueryType
  mutation: MyMutationType
}
```
````

jonathan-vooma: What percent of users would you estimate rename their entry points?

jonathanberger: What percent of users would you estimate rename their entry points?

benjie: Hard to say; but our very own `swapi-graphql` example API does: https://graphql.org/swapi-graphql/?query=%7B__typename%7D

benjie:
@jonathanberger I still would like your feedback on that last point, but I've gone ahead and committed the changes since there is overlapping work in #1800

jonathan-vooma:
I think your latest approach is an improvement. Thanks!

jonathanberger:
I think your latest approach is an improvement. Thanks!

benjie:
Awesome; your work has been incorporated into https://github.com/graphql/graphql.github.io/pull/1800 so I'm going to close this PR - thanks again @jonathanberger :raised_hands:

PR #1765, opened 2024-08-16
Add new python client library `ql`

dsal3389:
Hi, I would like to add graphql client library I am currently working on `ql`.

PR #1796, opened 2024-10-16
Label mini GraphiQL panels.

### LINKED ISSUE #250: Documentation about adding a mutation was hard to follow
I was looking into how to add a "Create User" mutation to my graphql-js server. It took me a while to end up at the right code, so just documenting where I got confused:

1. Searched for "Mutation" on the site. There are two sections ostensibly about mutation:
 a. http://graphql.org/learn/queries/#mutations
  b. http://graphql.org/learn/schema/#the-query-and-mutation-types
1. Skimming each page, it was hard to figure out how to actually define a mutation in the schema. The code talking about mutations wasn't clear if it was showing how to do them in the query or in the schema.
1. I found the operative sentence is in (b): "Mutations work in a similar way - you define fields on the Mutation type, and those are available as the root mutation fields you can call in your query."  and was able to write the correct schema:

```
  input CreatePersonInput {
    name: String
  }
  type Mutation {
    createPerson(person: CreatePersonInput!): Person
  }
```

Suggestions:
- I was thinking it would be better to just put an example of the schema changes necessary to make a mutation right into the schema (b) section to avoid confusion
- Make it clearer when a block of code in the docs is in GraphQL query language or schema language (maybe even show "Query" and "Response" labels next to the left and right sides)

mandiwise:
Partially addresses #250 (addresses the second bullet point, a subsequent PR will address the first)

## Description

This PR adds labels to each panel in the mini GraphiQL editors to help disambiguate the different elements of the code examples (Operation, Variables, and Response).

### DIFF
--- a/src/components/marked/mini-graphiQL.tsx
+++ b/src/components/marked/mini-graphiQL.tsx
@@ -311,7 +311,7 @@ class ResultViewer extends Component {
   render() {
     return (
       <div className="result-window" ref={e => (this.domNode = e)}>
-        <span className="editor-name">Response</span>
+        <span className="editor-name rounded-tr">Response</span>
       </div>
     )
   }
--- a/src/components/marked/mini-graphiQL.tsx
+++ b/src/components/marked/mini-graphiQL.tsx
@@ -256,7 +256,11 @@ class QueryEditor extends Component {
   }
 
   render() {
-    return <div className="query-editor" ref={e => (this.domNode = e)} />
+    return (
+      <div className="query-editor" ref={e => (this.domNode = e)}>
+        <span className="editor-name">Operation</span>
+      </div>
+    )
   }
 }
 
@@ -305,7 +309,11 @@ class ResultViewer extends Component {
   }
 
   render() {
-    return <div className="result-window" ref={e => (this.domNode = e)} />
+    return (
+      <div className="result-window" ref={e => (this.domNode = e)}>
+        <span className="editor-name">Response</span>
+      </div>
+    )
   }
 }
 
@@ -424,7 +432,11 @@ class VariableEditor extends Component {
   }
 
   render() {
-    return <div className="variable-editor" ref={e => (this.domNode = e)} />
+    return (
+      <div className="variable-editor" ref={e => (this.domNode = e)}>
+        <span className="editor-name">Variables</span>
+      </div>
+    )
   }
 
   _didLint(annotations) {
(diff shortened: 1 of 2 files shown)


src/components/marked/mini-graphiQL.tsx
```diff
@@ -256,7 +256,11 @@ class QueryEditor extends Component {
   }
 
   render() {
-    return <div className="query-editor" ref={e => (this.domNode = e)} />
+    return (
+      <div className="query-editor" ref={e => (this.domNode = e)}>
+        <span className="editor-name">Operation</span>
```
benjie: Technically this is a "document" since it may contain fragments and multiple operations; but I think that's probably a little too technical for people learning the technology so I think you've made the right choice of term here :+1:

benjie (approved):
Nice improvement :+1:

dimaMachina (commented):
just a little suggestion, can you round borders? 
image
image

src/components/marked/mini-graphiQL.tsx
dimaMachina:
```suggestion
        <span className="editor-name rounded-tl">Operation</span>
```

src/components/marked/mini-graphiQL.tsx
```diff
@@ -305,7 +309,11 @@ class ResultViewer extends Component {
   }
 
   render() {
-    return <div className="result-window" ref={e => (this.domNode = e)} />
+    return (
+      <div className="result-window" ref={e => (this.domNode = e)}>
+        <span className="editor-name">Response</span>
```
dimaMachina:
```suggestion
        <span className="editor-name rounded-tr">Response</span>
```

dimaMachina:
@benjie feel free to apply my suggestions to make it perfect ;) 
image

Urigo:
I'm not sure if this PR is the reason, but the descriptions of the mini graphiql do not disappear ones shown:
image

benjie:
Doesn't seem to be related; I've looked at [the preview](https://graphql-github-io-git-fork-dsal3-c9de1e-the-graph-ql-foundation.vercel.app/learn/queries/) of https://github.com/graphql/graphql.github.io/pull/1765 which predates this change and it has the same issue. I think it's been an issue for a _long_ time.

dimaMachina:
Yeah, we need improve mini graphiql at some time

Issue #1698, opened 2024-05-10
Feedback for “Introduction to GraphQL”


IanThomwohl:
Your first phrase in your first sentence is "GraphQL is a query language for your API" but you never explain how to map the Query to the API. What am I missing?

IanThomwohl:
I think I see, it some sort of 'magic' in the resolvers. I have not seen how the magic happens yet.

benjie:
Please keep this stream of consciousness going as you explore the docs, it's valuable insight for us :raised_hands:

Issue #1112, opened 2021-09-13
Adding a quick link to the source code / schema of example at docs 


htbkoo:
Hi team,

I am trying to learn `GraphQL` by reading [the docs](https://graphql.org/).

While I personally found the [interactive examples](https://graphql.org/learn/queries/#fields) quite useful, I was also quite curious about the actual schema / data behind the interactive examples.

As discussed in #728, I believe [the schema and data can be found here at `graphql.github.io/site/_core/swapiSchema.js`](https://github.com/graphql/graphql.github.io/blob/739a1c913fa4a88ac7c64c93ce68fc64ec0cbd96/site/_core/swapiSchema.js#L21)

However, I wonder if we could make the documentation more user-friendly by adding a quick link at the top of the documentation page, say, something like `You can find the schema and data behind these interactive examples here`.

Anyway, thanks again for your tremendous efforts in maintaining the library and docs!

ThanushaMandala:
Hi Team,

the documentation seems to be a great place to start for GraphQL beginners, but having the access to schema and data as mentioned by @htbkoo would be really accelerating the understanding of the learners.

Thank you for building an amazing resource for GraphQL enthusiasts :)

PR #1799, opened 2024-10-20
Update Learn introduction page.

mandiwise:
Addresses #1112 and #1698

## Description

This PR contains revised content of for the Introduction page of the Learn section.

@jorydotcom @benjie

### DIFF
--- a/src/globals.css
+++ b/src/globals.css
@@ -486,3 +486,11 @@ div[id^="headlessui-menu-items"] {
   font-size: 13px;
   padding: 7px 14px;
 }
+
+.learn-subtitle {
+  @apply mt-2;
+  @apply pb-2;
+  @apply border-neutral-200;
+  @apply border-b;
+  @apply text-xl;
+}
--- a/src/globals.css
+++ b/src/globals.css
@@ -493,4 +493,5 @@ div[id^="headlessui-menu-items"] {
   @apply border-neutral-200;
   @apply border-b;
   @apply text-xl;
+  @apply dark:border-neutral-700/80;
 }
--- a/src/pages/learn/index.mdx
+++ b/src/pages/learn/index.mdx
@@ -4,11 +4,11 @@ import { Callout } from "nextra/components"
 
 <p className="learn-subtitle">Learn about GraphQL, how it works, and how to use it</p>
 
-GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your data. The [GraphQL specification](https://spec.graphql.org/) was open-sourced in 2015 and has since been implemented in a variety of programming languages. GraphQL also isn't tied to any specific database or storage engine and is instead backed by your existing code and data.
+GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your data. The [GraphQL specification](https://spec.graphql.org/) was open-sourced in 2015 and has since been implemented in a variety of programming languages. GraphQL isn't tied to any specific database or storage engine—it is backed by your existing code and data.
 
 ## Describe your API with a type system
 
-A GraphQL service is created by [defining types and their fields](/learn/schema/), and then writing functions that [provide data for each field](/learn/execution/). For example, a GraphQL service that tells you the name of a logged-in user might look like this:
+A GraphQL service is created by [defining types and their fields](/learn/schema/), and then writing a function for each field to [provide the required data](/learn/execution/). For example, a GraphQL service that tells you the name of a logged-in user might look like this:
 
 ```graphql
 type Query {
@@ -78,7 +78,7 @@ type User {
 }
 ```
 
-GraphQL will continue to provide data as expected for the deprecated `name` field until it's determined that clients have updated their existing queries and the older field can be safely removed.
+Client tooling will encourage developers to use the new fields and remove usage of the deprecated `name` field. The field can be removed once it is determined it is no longer used; in the meantime GraphQL will continue to provide its data as expected.
 
 ## Try it out!
 
--- a/src/pages/learn/index.mdx
+++ b/src/pages/learn/index.mdx
@@ -26,17 +26,17 @@ Along with functions for each field on each type:
 
 ```js
 // Provide data for the `me` field on the `Query` type
-function Query_me(obj, args, context, info) {
+function Query_me(query, args, context, info) {
   return context.request.auth.user
 }
 
 // Provide data for the `name` field on the `User` type
-function User_name(obj, args, context, info) {
-  return context.db.getUserFullName(obj.id)
+function User_name(user, args, context, info) {
+  return context.db.getUserFullName(user.id)
 }
 ```
 
-In the example above, the function that provides data for the `me` field on the `Query` type uses information about the authenticated user who made the request, while the the `name` field on the `User` type is populated by using that user's ID to fetch their full name from an existing database.
+In the example above, the function that provides data for the `me` field on the `Query` type uses information about the authenticated user who made the request, while the the `name` field on the `User` type is populated by using that user's ID to fetch their full name from a database.
 
 ## Query exactly what you need
 
@@ -64,7 +64,7 @@ Could produce the following JSON result:
 }
 ```
 
-With even a simple query, we can see some of the key features that make GraphQL so powerful. The client can make queries to the API that mirror the structure of the data that they need and then receive just that data in the expected shape with a single request—and without having to worry about what underlying data sources provided it.
+With even a simple query, we can see some of the key features that make GraphQL so powerful. The client can make queries to the API that mirror the structure of the data that they need and then receive just that data in the expected shape with a single request—and without having to worry about which underlying data sources provided it.
 
 ## Evolve your API without versioning
 
--- a/src/pages/learn/index.mdx
+++ b/src/pages/learn/index.mdx
@@ -1,15 +1,14 @@
+import { Callout } from "nextra/components"
+
 # Introduction to GraphQL
 
-> Learn about GraphQL, how it works, and how to use it. Looking for documentation on how to build a GraphQL service?
-> There are libraries to help you implement GraphQL in [many different languages](/code). For an in-depth learning experience
-> with practical tutorials, see [the available training courses](/community/resources/training-courses).
+<p className="learn-subtitle">Learn about GraphQL, how it works, and how to use it</p>
 
-GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your
-data. GraphQL isn't tied to any specific database or storage engine and is instead backed by your existing code and data.
+GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your data. The [GraphQL specification](https://spec.graphql.org/) was open-sourced in 2015 and has since been implemented in a variety of programming languages. GraphQL also isn't tied to any specific database or storage engine and is instead backed by your existing code and data.
 
-A GraphQL service is created by defining types and fields on those types, then providing functions for each field on each type.
-For example, a GraphQL service that tells you who the logged in user is (`me`) as well as that user's name might look
-like this:
+## Describe your API with a type system
+
+A GraphQL service is created by [defining types and their fields](/learn/schema/), and then writing functions that [provide data for each field](/learn/execution/). For example, a GraphQL service that tells you the name of a logged-in user might look like this:
 
 ```graphql
 type Query {
@@ -17,7 +16,6 @@ type Query {
 }
 
 type User {
-  id: ID
   name: String
 }
 ```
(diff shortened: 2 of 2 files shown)


src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
+import { Callout } from "nextra/components"
+
 # Introduction to GraphQL
 
-> Learn about GraphQL, how it works, and how to use it. Looking for documentation on how to build a GraphQL service?
-> There are libraries to help you implement GraphQL in [many different languages](/code). For an in-depth learning experience
-> with practical tutorials, see [the available training courses](/community/resources/training-courses).
+<p className="learn-subtitle">Learn about GraphQL, how it works, and how to use it</p>
+
+GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your data. The [GraphQL specification](https://spec.graphql.org/) was open-sourced in 2015 and has since been implemented in a variety of programming languages. GraphQL also isn't tied to any specific database or storage engine and is instead backed by your existing code and data.
```
benjie:
```suggestion
GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your data. The [GraphQL specification](https://spec.graphql.org/) was open-sourced in 2015 and has since been implemented in a variety of programming languages. GraphQL isn't tied to any specific database or storage engine—it is backed by your existing code and data.
```

src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
...
 
-GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your
-data. GraphQL isn't tied to any specific database or storage engine and is instead backed by your existing code and data.
+## Describe your API with a type system
 
-A GraphQL service is created by defining types and fields on those types, then providing functions for each field on each type.
-For example, a GraphQL service that tells you who the logged in user is (`me`) as well as that user's name might look
-like this:
+A GraphQL service is created by [defining types and their fields](/learn/schema/), and then writing functions that [provide data for each field](/learn/execution/). For example, a GraphQL service that tells you the name of a logged-in user might look like this:
```
benjie: This could be interpreted as you write a few functions that each apply to many fields; though there are functions like this (e.g. the default resolver), most fields get their own function. This slight rewording hopefully removes this ambiguity?

```suggestion
A GraphQL service is created by [defining types and their fields](/learn/schema/), and then writing a function for each field to [provide the required data](/learn/execution/). For example, a GraphQL service that tells you the name of a logged-in user might look like this:
```

src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
...
   name: String
 }
 ```
 
 Along with functions for each field on each type:
 
 ```js
-function Query_me(request) {
-  return request.auth.user
+Query: {
```
benjie: This is no longer standalone JS code (`Query` here would be interpreted as a [label](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label) which could be very confusing for readers new to JavaScript), and it ties into the `{[typeName]: {[fieldName]: resolver}}` pattern common in various tools across the JavaScript ecosystem but not actually supported by graphql.js natively and also not as portable to other programming languages.

`function Query_me(request)` has the advantage of being fairly generic pseudocode even if the reader does not know JavaScript.

Perhaps adding a comment before the function would be a suitable alternative solution:

```ts
// The resolver for the `me` field on the `Query` type, aka `Query.me`:
function Query_me(request) {
```

and later:

```ts
// The resolver for the `User.name` field:
function User_name(user) {
```

what do you think?

src/pages/learn/index.mdx
```diff
@@ -52,8 +56,52 @@ Could produce the following JSON result:
...
+
+```graphql
+type User {
+  fullName: String
+  nickname: String
+  name: String @deprecated(reason: "Use `fullName`.")
+}
+```
+
+GraphQL will continue to provide data as expected for the deprecated `name` field until it's determined that clients have updated their existing queries and the older field can be safely removed.
```
benjie: This edit clarifies that releasing an update to all of the clients is not sufficient - we also need to ensure that no existing (non-updated) clients remain - e.g. mobile apps from years ago. It also makes it clear that GraphQL itself does not make any determination that the field is no longer used.

```suggestion
Client tooling will encourage developers to use the new fields and remove usage of the deprecated `name` field. The field can be removed once it is determined it is no longer used; in the meantime GraphQL will continue to provide its data as expected.
```

src/pages/learn/index.mdx
```diff
@@ -52,8 +56,52 @@ Could produce the following JSON result:
...
+
+<Callout type="info">
+The examples in this guide are based on [a modified version of the SWAPI GraphQL schema](https://github.com/graphql/graphql.github.io/blob/source/src/components/marked/swapi-schema.tsx). Because these queries are designed for illustrative purposes, they will not run on the full version of the SWAPI GraphQL API due to differences between the two schemas. [You can try the full version of the API here.](https://graphql.org/swapi-graphql/)
+</Callout>
+
+## Next steps
+
+Now that you know some key GraphQL concepts, you're ready to learn about all of the different aspects of its [type system](/schema/).
+
+If you're already familiar with GraphQL and would like to read documentation on how to build a GraphQL service, then there are libraries to help you implement GraphQL in [many different languages](/community/tools-and-libraries/?tags=server). There are also many libraries available that allow [client applications to query existing GraphQL APIs](/community/tools-and-libraries/?tags=client).
```
benjie: If someone is already familiar with GraphQL it seems unlikely they would read this far - perhaps this should be in a callout at the top of the page? I'm not super familiar with Nextra, but in Docusaurus I'd use an "info" [admonition](https://docusaurus.io/docs/markdown-features/admonitions).

benjie (changes requested):
This is great, some really good improvements here! I've made some suggested edits where I felt there was some potential ambiguity in the wording in a couple places, or where the copy could be tightened up slightly; let me know what you think.

src/pages/learn/index.mdx
mandiwise: I see what you mean with the label confusion. My intent here was to standardize these examples with the JS-ish examples that were already on the Execution page. But maybe a `function {Type}_{field} () {}` syntax would be best to use throughout the entire guide instead? Let me know what you think and I can adjust them.

```js
// Provide data for the `me` field on the `Query` type
function Query_me(obj, args, context) {
  return context.request.auth.user
}

// Provide data for the `name` field on the `User` type
function User_name(obj, args, context) {
  return context.db.getUserFullName(obj.id)
}
```

src/pages/learn/index.mdx
mandiwise: Ok, I'll move this back to the top.

src/pages/learn/index.mdx
benjie: Urgh, yes, how inconsistent! I do like the structure of typeName/fieldName for resolver clarity but the invalid JS code is not helpful. Out of interest I fed the code from the execution page into ChatGPT to see what it made of it:

Converting the invalid JS code to Python:

https://chatgpt.com/share/671a2803-2338-8012-aaee-c7eb490b72ce (doesn't look useful)

Converting expanded valid JS code to Python:

https://chatgpt.com/share/671a2843-3f08-8012-83d7-8769c471ecd9 (looks much more reasonable to me).

Disclaimer: I've not written Python for ~8 years.

I think I'd stick with functions like you suggest above on this page, and then on the execution page if it makes sense to do the typeName/fieldName structure, explicitly introduce it (e.g. "for organizational reasons, you may wish to keep your resolvers in a structured object, for example ...")

mandiwise: Sounds good. I updated the functions as discussed and moved the libraries callout in the latest commit so this should resolve all of the feedback for this page.

src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
...
+// Provide data for the `me` field on the `Query` type
+function Query_me(obj, args, context, info) {
+  return context.request.auth.user
 }
 
-function User_name(user) {
-  return user.getName()
+// Provide data for the `name` field on the `User` type
+function User_name(obj, args, context, info) {
+  return context.db.getUserFullName(obj.id)
```
benjie:
```suggestion
function User_name(user, args, context, info) {
  return context.db.getUserFullName(user.id)
```

src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
...
 ```
 
 Along with functions for each field on each type:
 
 ```js
-function Query_me(request) {
-  return request.auth.user
+// Provide data for the `me` field on the `Query` type
+function Query_me(obj, args, context, info) {
+  return context.request.auth.user
```
benjie: I wasn't sure whether to put `query` or `rootValue` or `source` or `parent` or `_` here; I went with `query` in the end for consistency with `user` below.

```suggestion
function Query_me(query, args, context, info) {
  return context.request.auth.user
```

src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
...
+// Provide data for the `name` field on the `User` type
+function User_name(obj, args, context, info) {
+  return context.db.getUserFullName(obj.id)
 }
 ```
 
-After a GraphQL service is running (typically at a URL on a web service), it can receive GraphQL queries to validate and execute.
-The service first checks a query to ensure it only refers to the types and fields defined, and then runs the provided functions
-to produce a result.
+In the example above, the function that provides data for the `me` field on the `Query` type uses information about the authenticated user who made the request, while the the `name` field on the `User` type is populated by using that user's ID to fetch their full name from an existing database.
```
benjie: I don't think "existing" is an important qualifier here, unless I'm missing something?

```suggestion
In the example above, the function that provides data for the `me` field on the `Query` type uses information about the authenticated user who made the request, while the the `name` field on the `User` type is populated by using that user's ID to fetch their full name from a database.
```

src/pages/learn/index.mdx
```diff
@@ -52,8 +56,50 @@ Could produce the following JSON result:
...
-    "name": "Luke Skywalker"
+  "data": {
+    "me": {
+      "name": "Luke Skywalker"
+    }
   }
 }
 ```
+
+With even a simple query, we can see some of the key features that make GraphQL so powerful. The client can make queries to the API that mirror the structure of the data that they need and then receive just that data in the expected shape with a single request—and without having to worry about what underlying data sources provided it.
```
benjie:
```suggestion
With even a simple query, we can see some of the key features that make GraphQL so powerful. The client can make queries to the API that mirror the structure of the data that they need and then receive just that data in the expected shape with a single request—and without having to worry about which underlying data sources provided it.
```

benjie (changes requested):
This is great! :star_struck:

benjie (approved):

benjie:
I've applied some minor edits; I'll leave this open for a few days to see if there's any other feedback but I think it's good to merge - thanks for your work on this, it's a great improvement in my opinion!

Urigo (approved):

src/globals.css
```diff
@@ -486,3 +486,11 @@ div[id^="headlessui-menu-items"] {
   font-size: 13px;
   padding: 7px 14px;
 }
+
+.learn-subtitle {
+  @apply mt-2;
+  @apply pb-2;
+  @apply border-neutral-200;
+  @apply border-b;
```
dimaMachina, line 494: border color should be improved on dark mode 
image

src/pages/learn/index.mdx
```diff
@@ -1,42 +1,46 @@
...
+<p className="learn-subtitle">Learn about GraphQL, how it works, and how to use it</p>
+
+GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your data. The [GraphQL specification](https://spec.graphql.org/) was open-sourced in 2015 and has since been implemented in a variety of programming languages. GraphQL isn't tied to any specific database or storage engine—it is backed by your existing code and data.
 
-GraphQL is a query language for your API, and a server-side runtime for executing queries using a type system you define for your
-data. GraphQL isn't tied to any specific database or storage engine and is instead backed by your existing code and data.
+<Callout type="info">If you're already familiar with GraphQL and would like to read documentation on how to build a GraphQL service, then there are libraries to help you implement GraphQL in [many different languages](/community/tools-and-libraries/?tags=server). There are also many libraries available that allow [client applications to query existing GraphQL APIs](/community/tools-and-libraries/?tags=client).</Callout>
```
saihaj, line 9: Should we mention about GraphQL.js being the official reference implementation maintained by the GraphQL foundation?

saihaj (approved):
looks amazing!

src/globals.css
mandiwise: Border color for dark mode has been adjusted.

Screenshot 2024-10-29 at 2 51 01 PM

src/pages/learn/index.mdx
mandiwise: @benjie Do you want to call out graphql.js specifically here?

benjie: I don't think we need to do that here, but our first mention of a "reference implementation" should hyperlink to GraphQL.js

benjie:
Excellent work @mandiwise and great feedback everyone :raised_hands:

PR #1800, opened 2024-10-23
Update schema and types docs

mandiwise:
## Description

This PR contains updated content for the Schema and Types page in the Learn docs. Key changes include:

- An updated intro and a new "next steps" section at the end of the page
- A new section on type system directives
- A new section on documentation (descriptions and comments)
- Updated/expanded content and examples for all existing sections
- Standardization of capitalization and naming for named types

Note that the intention is to swap the order of this page in the left nav so that it comes before the Queries and Mutations content, but this will be addressed in the next PR that updates that operation-related content.

@benjie @jorydotcom

### DIFF
--- a/src/pages/learn/schema.mdx
+++ b/src/pages/learn/schema.mdx
@@ -165,7 +165,7 @@ This means that wherever we use the type `Episode` in our schema, we expect it t
 
 ## Type modifiers
 
-Types are assumed to be nullable and singular by default in GraphQL. However, when you use these named types in a schema (or in [query variable declarations](/learn/queries/#variables)) you can apply additional _type modifiers_ that will affect the validation of those values.
+Types are assumed to be nullable and singular by default in GraphQL. However, when you use these named types in a schema (or in [query variable declarations](/learn/queries/#variables)) you can apply additional _type modifiers_ that will affect the meaning of those values.
 
 As we saw with the Object type example above, GraphQL supports two type modifiers—the [List](https://spec.graphql.org/October2021/#sec-List) and [Non-Null](https://spec.graphql.org/October2021/#sec-List) types—and they can be used individually or in combination with each other.
 
@@ -179,7 +179,7 @@ type Character {
 }
 ```
 
-Here, we're using a `String` type and marking it as a Non-Null type by adding an exclamation mark (`!`) after the type name. This means that our server always expects to return a non-null value for this field, and if it receives a null value, then that will trigger a GraphQL execution error, letting the client know that something has gone wrong.
+Here, we're using a `String` type and marking it as a Non-Null type by adding an exclamation mark (`!`) after the type name. This means that our server always expects to return a non-null value for this field, and if the resolver produces a null value, then that will trigger a GraphQL execution error, letting the client know that something has gone wrong.
 
 As we saw in an example above, the Non-Null type modifier can also be used when defining arguments for a field, which will cause the GraphQL server to return a validation error if a null value is passed as that argument:
 
--- a/src/pages/learn/schema.mdx
+++ b/src/pages/learn/schema.mdx
@@ -4,9 +4,9 @@ import { Callout } from "nextra/components"
 
 <p className="learn-subtitle">Learn about the different elements of the GraphQL type system</p>
 
-The GraphQL [type system](https://spec.graphql.org/October2021/#sec-Type-System) describes what data can be queried from the API. The collection of those capabilities is referred to as the service's _schema_ and clients can use that schema to send queries to the API that return predictable results.
+The GraphQL [type system](https://spec.graphql.org/draft/#sec-Type-System) describes what data can be queried from the API. The collection of those capabilities is referred to as the service's _schema_ and clients can use that schema to send queries to the API that return predictable results.
 
-On this page, we'll explore GraphQL's [six kinds of named type definitions](https://spec.graphql.org/October2021/#sec-Types) as well as other features of the type system to learn how they may be used to describe your data and the relationships between them. Since GraphQL can be used with any backend framework or programming language, we'll avoid implementation-specific details and talk only about the concepts.
+On this page, we'll explore GraphQL's [six kinds of named type definitions](https://spec.graphql.org/draft/#sec-Types) as well as other features of the type system to learn how they may be used to describe your data and the relationships between them. Since GraphQL can be used with any backend framework or programming language, we'll avoid implementation-specific details and talk only about the concepts.
 
 ## Type system
 
@@ -38,7 +38,7 @@ Since we can't rely on a specific programming language to discuss GraphQL schema
 
 ## Object types and fields
 
-The most basic components of a GraphQL schema are [Object types](https://spec.graphql.org/October2021/#sec-Objects), which just represent a kind of object you can fetch from your service, and what fields it has. In SDL, we represent it like this:
+The most basic components of a GraphQL schema are [Object types](https://spec.graphql.org/draft/#sec-Objects), which just represent a kind of object you can fetch from your service, and what fields it has. In SDL, we represent it like this:
 
 ```graphql
 type Character {
@@ -59,7 +59,7 @@ Now you know what a GraphQL Object type looks like and how to read the basics of
 
 ### Arguments
 
-Every field on a GraphQL Object type can have zero or more [arguments](https://spec.graphql.org/October2021/#sec-Field-Arguments), for example, the `length` field below:
+Every field on a GraphQL Object type can have zero or more [arguments](https://spec.graphql.org/draft/#sec-Field-Arguments), for example, the `length` field below:
 
 ```graphql
 type Starship {
@@ -75,7 +75,7 @@ Arguments can be either required or optional. When an argument is optional, we c
 
 ### The Query, Mutation, and Subscription types
 
-Every GraphQL schema must support `query` operations. The _entry point_ for this [root operation type](https://spec.graphql.org/October2021/#sec-Root-Operation-Types) is a regular Object type called `Query` by default. So if you see a query that looks like this:
+Every GraphQL schema must support `query` operations. The _entry point_ for this [root operation type](https://spec.graphql.org/draft/#sec-Root-Operation-Types) is a regular Object type called `Query` by default. So if you see a query that looks like this:
 
 ```graphql
 # { "graphiql": true }
@@ -110,7 +110,7 @@ schema {
 
 ## Scalar types
 
-A GraphQL Object type has a name and fields, but at some point, those fields must resolve to some concrete data. That's where the [Scalar types](https://spec.graphql.org/October2021/#sec-Scalars) come in: they represent the leaf values of the query.
+A GraphQL Object type has a name and fields, but at some point, those fields must resolve to some concrete data. That's where the [Scalar types](https://spec.graphql.org/draft/#sec-Scalars) come in: they represent the leaf values of the query.
 
 In the following query, the `name` and `appearsIn` fields will resolve to Scalar types:
 
@@ -126,7 +126,7 @@ In the following query, the `name` and `appearsIn` fields will resolve to Scalar
 
 We know this because those fields don't have any sub-fields—they are the leaves of the query.
 
-GraphQL comes with a set of [default Scalar types](https://spec.graphql.org/October2021/#sec-Scalars.Built-in-Scalars) out of the box:
+GraphQL comes with a set of [default Scalar types](https://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars) out of the box:
 
 - `Int`: A signed 32‐bit integer.
 - `Float`: A signed double-precision floating-point value.
@@ -144,7 +144,7 @@ Then it's up to our implementation to define how that type should be serialized,
 
 ## Enum types
 
-[Enum types](https://spec.graphql.org/October2021/#sec-Enums), also known as enumeration types, are a special kind of scalar that is restricted to a particular set of allowed values. This allows you to:
+[Enum types](https://spec.graphql.org/draft/#sec-Enums), also known as enumeration types, are a special kind of scalar that is restricted to a particular set of allowed values. This allows you to:
 
 1. Validate that any arguments of this type are one of the allowed values
 2. Communicate through the type system that a field will always be one of a finite set of values
@@ -167,7 +167,7 @@ This means that wherever we use the type `Episode` in our schema, we expect it t
 
 Types are assumed to be nullable and singular by default in GraphQL. However, when you use these named types in a schema (or in [query variable declarations](/learn/queries/#variables)) you can apply additional _type modifiers_ that will affect the meaning of those values.
 
-As we saw with the Object type example above, GraphQL supports two type modifiers—the [List](https://spec.graphql.org/October2021/#sec-List) and [Non-Null](https://spec.graphql.org/October2021/#sec-List) types—and they can be used individually or in combination with each other.
+As we saw with the Object type example above, GraphQL supports two type modifiers—the [List](https://spec.graphql.org/draft/#sec-List) and [Non-Null](https://spec.graphql.org/draft/#sec-Non-Null) types—and they can be used individually or in combination with each other.
 
 ### Non-Null
 
@@ -254,7 +254,7 @@ You can arbitrarily nest any number of Non-Null and List modifiers, according to
 
 ## Interface types
 
-Like many type systems, GraphQL supports _abstract types_. The first of these types that we'll explore is an [Interface type](https://spec.graphql.org/October2021/#sec-Interfaces), which defines a certain set of fields that a concrete Object type or other Interface type must also include to implement it.
+Like many type systems, GraphQL supports _abstract types_. The first of these types that we'll explore is an [Interface type](https://spec.graphql.org/draft/#sec-Interfaces), which defines a certain set of fields that a concrete Object type or other Interface type must also include to implement it.
 
 For example, you could have an `Character` Interface type that represents any character in the Star Wars trilogy:
 
@@ -341,7 +341,7 @@ Note that Interface types may not implement themselves or contain any cyclic ref
 
 ## Union types
 
-GraphQL supports a second abstract type called a [Union type](https://spec.graphql.org/October2021/#sec-Unions). They share similarities with Interface types, but they cannot define any shared fields among the constituent types.
+GraphQL supports a second abstract type called a [Union type](https://spec.graphql.org/draft/#sec-Unions). They share similarities with Interface types, but they cannot define any shared fields among the constituent types.
 
 A Union type is defined by indicating its member Object types:
 
@@ -406,9 +406,9 @@ Note that `name` is still specified on `Starship` because otherwise it wouldn't 
 
 Most of the examples we've covered on this page demonstrate how Object, Scalar, Enum, Interface, and Union types may be used as _output types_ for the fields in a schema. But we have also seen that field arguments must specify their _input types_. 
 
-So far, we've only talked about using scalar values (like Enums or String types) as an input type for a field argument. However, you can also easily pass complex objects as arguments using an [Input Object type](https://spec.graphql.org/October2021/#sec-Interfaces), which is the last of named types in GraphQL that we'll explore.
+So far, we've only talked about using scalar values (like Enums or String types) as an input type for a field argument. However, you can also easily pass complex objects as arguments using an [Input Object type](https://spec.graphql.org/draft/#sec-Input-Objects), which is the last of named types in GraphQL that we'll explore.
 
-This is particularly valuable in the case of [mutations](https://graphql.org/learn/mutations/), where you might want to pass in a whole object to be created. In SDL, Input Object types look the same as regular Object types, but with the keyword `input` instead of `type`:
+This is particularly valuable in the case of [mutations](/learn/queries/#mutations), where you might want to pass in a whole object to be created. In SDL, Input Object types look the same as regular Object types, but with the keyword `input` instead of `type`:
 
 ```graphql
 input ReviewInput {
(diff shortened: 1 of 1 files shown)


benjie:
If possible, please rebase these changes onto https://github.com/graphql/graphql.github.io/pull/1794

src/pages/learn/schema.mdx
```diff
@@ -156,35 +161,49 @@ enum Episode {
...
+<Callout type="info">GraphQL service implementations in various languages will have a language-specific way of dealing with Enum types. In languages that support enums as a first-class citizen, the implementation might take advantage of that; in a language like JavaScript with no enum support, these values might be internally mapped to a set of integers. However, these details don't leak out to the client, which can operate entirely in terms of the string names of the Enum type's values.</Callout>
 
-## Lists and Non-Null
+## Type modifiers
 
-Object types, scalars, and enums are the only kinds of types you can define in GraphQL. But when you use the types in other parts of the schema, or in your query variable declarations, you can apply additional _type modifiers_ that affect validation of those values. Let's look at an example:
+Types are assumed to be nullable and singular by default in GraphQL. However, when you use these named types in a schema (or in [query variable declarations](/learn/queries/#variables)) you can apply additional _type modifiers_ that will affect the validation of those values.
```
benjie: I know `validation` was the previous text... but saying something is a list of int rather than an int doesn't just change its validation, it changes its actual meaning, right?

```suggestion
Types are assumed to be nullable and singular by default in GraphQL. However, when you use these named types in a schema (or in [query variable declarations](/learn/queries/#variables)) you can apply additional _type modifiers_ that will affect the meaning of those values.
```

src/pages/learn/schema.mdx
```diff
@@ -156,35 +161,49 @@ enum Episode {
...
 
 ```graphql
 type Character {
   name: String!
-  appearsIn: [Episode]!
 }
 ```
 
-Here, we're using a `String` type and marking it as _Non-Null_ by adding an exclamation mark, `!` after the type name. This means that our server always expects to return a non-null value for this field, and if it ends up getting a null value that will actually trigger a GraphQL execution error, letting the client know that something has gone wrong.
+Here, we're using a `String` type and marking it as a Non-Null type by adding an exclamation mark (`!`) after the type name. This means that our server always expects to return a non-null value for this field, and if it receives a null value, then that will trigger a GraphQL execution error, letting the client know that something has gone wrong.
```
benjie: The field resolver receives arguments and produces a result, we're talking about the result here so thinking of "the field produces" (rather than "GraphQL receives from the field") I think would be helpful.

```suggestion
Here, we're using a `String` type and marking it as a Non-Null type by adding an exclamation mark (`!`) after the type name. This means that our server always expects to return a non-null value for this field, and if the resolver produces a null value, then that will trigger a GraphQL execution error, letting the client know that something has gone wrong.
```

benjie (commented):
I wonder if we should be linking to the draft spec rather than a specific release. Everything in the draft has passed the RFC process and barring some exceptional circumstances will make it into the next official release - it's the version that implementations should be following/implementing, and the links are more evergreen (linking to section titles should be fine, but linking to specific words within a paragraph is probably riskier!).

mandiwise:
> If possible, please rebase these changes onto #1794

@benjie The commits from `jonathanberger/patch-1` have been merged into this branch now with the conflicts resolved.

mandiwise:
> I wonder if we should be linking to the draft spec rather than a specific release. Everything in the draft has passed the RFC process and barring some exceptional circumstances will make it into the next official release - it's the version that implementations should be following/implementing, and the links are more evergreen (linking to section titles should be fine, but linking to specific words within a paragraph is probably riskier!).

@benjie I'm happy to swap out the URLs to use the draft spec path so they're more stable over time. The other option I can think of would be to create something like a `<SpecLink section="some-section-id">` component to use throughout the MDX files so that when a new release of the spec is finalized the URL will only need to be updated in the component file. Let me know what you think.

benjie:
The MDX component is an interesting approach! I think we should just link to the draft - our general advice to everyone is to follow the draft, and the draft is where any typo fixes and other such things will go so it's generally more correct/up-to-date.

mandiwise:
Spec links have now been changed to the draft path.

Urigo (approved):

saihaj (approved):

src/pages/learn/schema.mdx
```diff
@@ -20,17 +26,19 @@ If you've seen a GraphQL query before, you know that the GraphQL query language
...
+That's where the schema comes in. Every GraphQL service defines a set of types that completely describe the set of possible data we can query on that service. Then, when requests come in, they are validated and executed against that schema.
 
 ## Type language
 
-GraphQL services can be written in any language. Since we can't rely on a specific programming language syntax, like JavaScript, to talk about GraphQL schemas, we'll define our own simple language. We'll use the "GraphQL schema language" - it's similar to the query language, and allows us to talk about GraphQL schemas in a language-agnostic way.
+GraphQL services can be written in any language and there are different approaches you can take when defining the types in a schema. Some libraries require that you only write the _resolver functions_ that provide data for the different fields in the API, then compile the schema based on that code. Other libraries allow you to define types and fields using what's commonly called the _schema definition language_ (or SDL) and then write the resolver functions for the corresponding fields separately.
```
benjie: Many libraries support more than one way of defining schemas, and there are many different approaches than implementation-first or SDL-first. This edit attempts to capture more of the possibilities to whet the appetite. (I started with the way that graphql-js primarily supports - i.e. creating class instances.)

Do you think this is too much content though?

```suggestion
GraphQL services can be written in any language and there are many different approaches you can take when defining the types in a schema. Some libraries have you construct the schema types, fields, and resolver functions together through traditional code. Some libraries allow you to define types and fields more ergonomically using what's commonly called the _schema definition language_ (or SDL) and then write the _resolver functions_ for the corresponding fields separately. Some libraries allow you to write and annotate the resolver functions, then infer the schema from that. Some libraries may even infer both the types and resolvers for you, based on some underlying datasource(s).
```

src/pages/learn/schema.mdx
```diff
@@ -20,17 +26,19 @@ If you've seen a GraphQL query before, you know that the GraphQL query language
...
+
+Since we can't rely on a specific programming language to discuss GraphQL schemas in this guide, we'll use SDL because it's similar to the query language that we've seen so far and allows us to talk about GraphQL schemas in a language-agnostic way.
 
 ## Object types and fields
 
-The most basic components of a GraphQL schema are object types, which just represent a kind of object you can fetch from your service, and what fields it has. In the GraphQL schema language, we might represent it like this:
+The most basic components of a GraphQL schema are [Object types](https://spec.graphql.org/draft/#sec-Objects), which just represent a kind of object you can fetch from your service, and what fields it has. In SDL, we represent it like this:
```
benjie, line 46: Trivia: the spec never refers to "SDL" itself but instead [uses the term "IDL"](https://spec.graphql.org/draft/#sel-DAHJCAACCyBpge). That said, it uses the term "type system definition language" in a number of places, so I think the use of "SDL" is fine (though we generally see this as expanding to "_schema_ definition language", a phrase which also does not occur in the GraphQL spec).

src/pages/learn/schema.mdx
```diff
@@ -39,19 +47,19 @@ type Character {
...
+- `Character` is a GraphQL Object type, meaning it's a type with some fields. Most of the types in your schema will be Object types.
+- `name` and `appearsIn` are fields on the `Character` type. That means that `name` and `appearsIn` are the only fields that can appear in any part of a GraphQL query that operates on the `Character` type.
+- `String` is one of the built-in Scalar types. These are types that resolve to a single scalar value and can't have sub-selections in the query. We'll go over Scalar types more later.
+- `String!` means that the field is a Non-Null type, meaning the GraphQL service promises to give you a value whenever you query this field. In SDL, we represent those with an exclamation mark.
+- `[Episode!]!` represents an List type of `Episode` objects. When a List is Non-Null, you can always expect an array (with zero or more items) when you query the `appearsIn` field. In this case, since `Episode!` is also Non-Null within the list, you can always expect every item in the array to be an `Episode` object.
```
benjie: Do you not feel that the italic helps with clarifying which term it is that we're actually defining in our vocabulary?

src/pages/learn/schema.mdx
```diff
@@ -61,53 +69,50 @@ type Starship {
...
 type Query {
-  hero(episode: Episode): Character
   droid(id: ID!): Droid
 }
 ```
 
-Mutations work in a similar way - you define fields on the `Mutation` type, and those are available as the root mutation fields you can call in your query.
+Schemas may also support `mutation` and `subscription` operations by adding additional `Mutation` and `Subscription` types and then defining fields on the corresponding root operation types.
+
+It's important to remember that apart from the special status of being entry points into the schema, the `Query`, `Mutation`, and `Subscription` types are the same as any other GraphQL Object type, and their fields work exactly the same way.
```
benjie:
```suggestion
It's important to remember that other than the special status of being entry points into the schema, the `Query`, `Mutation`, and `Subscription` types are the same as any other GraphQL Object type, and their fields work exactly the same way.
```

src/pages/learn/schema.mdx
```diff
@@ -119,32 +124,32 @@ In the following query, the `name` and `appearsIn` fields will resolve to scalar
...
 
-GraphQL comes with a set of default scalar types out of the box:
+GraphQL comes with a set of [default Scalar types](https://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars) out of the box:
 
 - `Int`: A signed 32‐bit integer.
 - `Float`: A signed double-precision floating-point value.
 - `String`: A UTF‐8 character sequence.
 - `Boolean`: `true` or `false`.
-- `ID`: The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, defining it as an `ID` signifies that it is not intended to be human‐readable.
+- `ID`: This type represents a unique identifier, often used to refetch an object or as the key for a cache. The `ID` type is serialized in the same way as a `String`; however, defining it as an `ID` signifies that it is not intended to be human‐readable.
```
benjie:
```suggestion
- `ID`: A unique identifier, often used to refetch an object or as the key for a cache. The `ID` type is serialized in the same way as a `String`; however, defining it as an `ID` signifies that it is not intended to be human‐readable.
```

src/pages/learn/schema.mdx
```diff
@@ -156,35 +161,49 @@ enum Episode {
...
+# { "graphiql": true }
+{
+  droid(id: null) {
     name
   }
 }
 ```
 
-Lists work in a similar way: We can use a type modifier to mark a type as a `List`, which indicates that this field will return an array of that type. In the schema language, this is denoted by wrapping the type in square brackets, `[` and `]`. It works the same for arguments, where the validation step will expect an array for that value.
+### Lists
```
benjie: For consistency with Non-Null, this should be singular:

```suggestion
### List
```

src/pages/learn/schema.mdx
```diff
@@ -254,30 +290,28 @@ type Droid implements Character {
...
-For example, note that the following query produces an error:
+Interface types are useful when you want to return an object or set of objects, but those might be of several different types. For example, note that the following query produces an error:
 
 ```graphql
-# { "graphiql": true, "variables": { "ep": "JEDI" } }
-query HeroForEpisode($ep: Episode!) {
-  hero(episode: $ep) {
+# { "graphiql": true }
+{
+  hero(episode: JEDI) {
```
benjie, line 305: I've noticed you've removed variables from a number of the examples and I appreciate that this simplifies the examples, but I worry that this encourages people to inline values and leads them down the path of concatenation to form GraphQL queries (rather than using static queries). By using variables in our examples, I feel that we're modelling best practices and subtly encouraging users to do the same. What do you think?

src/pages/learn/schema.mdx
```diff
@@ -286,19 +320,38 @@ query HeroForEpisode($ep: Episode!) {
...
+```
+
+Note that Interface types may not implement themselves or contain any cyclic references to each other.
+
+<Callout type="info">While types may share common fields in a GraphQL API, that may not always warrant the use of an Interface type to enforce that those fields remain consistently named. Interface types are a powerful way to indicate shared behavior across the types that implement them. When used, they should be semantically meaningful to client developers, just like `Character` is a useful abstraction for humans and droids.</Callout>
 
 ## Union types
 
-Union types share similarities with interfaces; however, they lack the ability to define any shared fields among the constituent types.
+GraphQL supports a second abstract type called a [Union type](https://spec.graphql.org/draft/#sec-Unions). They share similarities with Interface types, but they cannot define any shared fields among the constituent types.
```
benjie:
```suggestion
GraphQL supports a second abstract type called a [Union type](https://spec.graphql.org/draft/#sec-Unions). Union types share similarities with Interface types, but they cannot define any shared fields among the constituent types.
```

src/pages/learn/schema.mdx
```diff
@@ -321,11 +374,12 @@ In this case, if you query a field that returns the `SearchResult` union type, y
 }
 ```
 
-The `__typename` field resolves to a `String` which lets you differentiate different data types from each other on the client.
+The `__typename` field is a special _meta-field_ that resolves to a `String` and lets you differentiate between data types on the client.
```
benjie: I like this addition, is it worth expanding a little?

```suggestion
The `__typename` field is a special _meta-field_ that automatically exists on every Object type and resolves to the name of that type, providing a way to differentiate between data types on the client.
```

src/pages/learn/schema.mdx
```diff
@@ -348,27 +402,158 @@ Also, in this case, since `Human` and `Droid` share a common interface (`Charact
 
 Note that `name` is still specified on `Starship` because otherwise it wouldn't show up in the results given that `Starship` is not a `Character`!
 
-## Input types
+## Input Object types
+
+Most of the examples we've covered on this page demonstrate how Object, Scalar, Enum, Interface, and Union types may be used as _output types_ for the fields in a schema. But we have also seen that field arguments must specify their _input types_. 
+
+So far, we've only talked about using scalar values (like Enums or String types) as an input type for a field argument. However, you can also easily pass complex objects as arguments using an [Input Object type](https://spec.graphql.org/draft/#sec-Input-Objects), which is the last of named types in GraphQL that we'll explore.
```
benjie: `-easily` because we should avoid judgement of difficulty

`+kind` for precision

```suggestion
So far, we've only talked about using scalar values (like Enums or String types) as an input type for a field argument. However, you can also pass complex objects as arguments using an [Input Object type](https://spec.graphql.org/draft/#sec-Input-Objects), which is the last kind of named types in GraphQL that we'll explore.
```

src/pages/learn/schema.mdx
```diff
@@ -348,27 +402,158 @@ Also, in this case, since `Human` and `Droid` share a common interface (`Charact
...
 
-So far, we've only talked about passing scalar values, like enums or strings, as arguments into a field. But you can also easily pass complex objects. This is particularly valuable in the case of mutations, where you might want to pass in a whole object to be created. In the GraphQL schema language, input types look exactly the same as regular object types, but with the keyword `input` instead of `type`:
+This is particularly valuable in the case of [mutations](/learn/queries/#mutations), where you might want to pass in a whole object to be created. In SDL, Input Object types look the same as regular Object types, but with the keyword `input` instead of `type`:
```
benjie: This is an issue with the previous text, but input objects don't accept arguments and can't be members of interfaces so they don't look exactly the same as objects, I think "similar to" would be more accurate?

```suggestion
This is particularly valuable in the case of [mutations](/learn/queries/#mutations), where you might want to pass in a whole object to be created. In SDL, Input Object types look similar to regular Object types, but with the keyword `input` instead of `type`:
```

src/pages/learn/schema.mdx
```diff
@@ -348,27 +402,158 @@ Also, in this case, since `Human` and `Droid` share a common interface (`Charact
...
+  """
+  hero(
+    "The name of the film that the hero appears in."
+    episode: Episode
+  ): Character
+}
+```
+
+In addition to making a GraphQL API schema more expressive, descriptions are helpful to client developers because they are available in [introspection queries](/learn/introspection/) and visible in developer tools such as [GraphiQL](https://github.com/graphql/graphiql).
+### Comments
```
benjie, line 525:
```suggestion

### Comments
```

benjie: Also it may be worth demonstrating that comments can come at the end of a line?

src/pages/learn/schema.mdx
```diff
@@ -348,27 +402,158 @@ Also, in this case, since `Human` and `Droid` share a common interface (`Charact
...
+
+## Next steps
+
+To recap what we've learned about schemas and types:
+
+- Depending on what library is selected to build a GraphQL service, we can define a schema in a language-agnostic way using SDL or by compiling the schema from the code that provides data for its fields
+- There are six kinds of named type definitions in GraphQL: Object, Scalar, Enum, Interface, Union, and Input Object types
+- Object types contain fields that specify output types, and those fields may have arguments that specify input types
+- The `Int`, `Float`, `String`, `Boolean`, and `ID` Scalar types are built into GraphQL, and you can also define your own custom scalars
+- Like Scalar types, Enum types represent leaf values in a GraphQL schema but they are limited to a finite list of values
```
benjie:
```suggestion
- Like Scalar types, Enum types represent leaf values in a GraphQL schema but they are limited to a finite set of values
```

src/pages/learn/schema.mdx
```diff
@@ -348,27 +402,158 @@ Also, in this case, since `Human` and `Droid` share a common interface (`Charact
...
+- Interface and Union types are abstract types that allow different concrete Object types to be output from a single field
+- Input Object types allow you to pass more complex values into a field argument than Scalar and Enum types allow
```
benjie:
```suggestion
- Input Object types allow you to pass more complex values into a field argument or directive argument than Scalar and Enum types allow
```

(I tried "field or directive argument" but that came across as ambiguous.)

src/pages/learn/schema.mdx
```diff
@@ -348,27 +402,158 @@ Also, in this case, since `Human` and `Droid` share a common interface (`Charact
...
+- Type system directives can be applied to the types, fields, and arguments in a schema to alter how they are validated and executed during a query
+- GraphQL supports schema documentation using type, field, and argument descriptions, and it also supports comments that are ignored by the parser
```
benjie, line 564: The summary seems to miss the type modifiers?

benjie (changes requested):
This is excellent! I've included some suggested edits, but I think this is a great improvement over the existing text - excellent work :raised_hands:

src/pages/learn/schema.mdx
mandiwise: I was on the fence about using "SDL" for this reason but I ended up introducing it on this page because it is so ubiquitous outside of the spec, so I thought it might be best for clarity's sake. Do you think it would be helpful to indicate that this is a common term (not a spec term) or is that too much information at this point?

src/pages/learn/schema.mdx
mandiwise: It felt like there was a lot going on visually with all of the code text and the italic formatting in the bullets, but maybe it makes sense to add anchor links around the terms to where they are defined in the sections below?

src/pages/learn/schema.mdx
mandiwise: My thinking here was that because #1802 swaps the order of this page in the left nav so that it comes before the operation-related content that it might be confusing to use the variable syntax here until that concept is introduced on the next page.

src/pages/learn/schema.mdx
mandiwise: Perhaps with this many options we should break it up into a bulleted list:

> GraphQL services can be written in any language and there are many different approaches you can take when defining the types in a schema:
> 
> - Some libraries have you construct the schema types, fields, and [resolver functions](/learn/execution/) together using the same language that was used to write the GraphQL implementation.
> - Some libraries allow you to define types and fields more ergonomically using what's commonly called the _schema definition language_ (or SDL) and then write the resolver functions for the corresponding fields separately.
> - Some libraries allow you to write and annotate the resolver functions, and then infer the schema from that.
> - Some libraries may even infer both the types and resolver functions for you, based on some underlying data source(s).

src/pages/learn/schema.mdx
mandiwise: Good catch, this has been fixed in a new commit.

src/pages/learn/schema.mdx
benjie: Sounds like a good idea! I'd change "language" to "programming language" in the first bullet.

src/pages/learn/schema.mdx
benjie: I think what you have is fine, truly this was just trivia. Thank you for thinking about it! (Perhaps the first time we use the term "SDL" we could says something like "colloquially known as SDL"? No big deal either way.)

src/pages/learn/schema.mdx
benjie: That sounds like a good compromise :+1:

src/pages/learn/schema.mdx
benjie: I'm okay with that; thank you for sharing the reasoning :+1:

benjie (changes requested):
Looking good just [the bulleted list of schema build methods](https://github.com/graphql/graphql.github.io/pull/1800#discussion_r1826255413) and maybe [the linking of the defined terms](https://github.com/graphql/graphql.github.io/pull/1800#discussion_r1826239212) left to consider I think!

mandiwise:
> Looking good just [the bulleted list of schema build methods](https://github.com/graphql/graphql.github.io/pull/1800#discussion_r1826255413) and maybe [the linking of the defined terms](https://github.com/graphql/graphql.github.io/pull/1800#discussion_r1826239212) left to consider I think!

These changes have been committed now.

benjie (approved):
Awesome; this is great! :raised_hands:

Issue #992, opened 2020-12-15
The learning section for Mutation is missing example for both Update and Upsert


StingyJack:
https://graphql.org/learn/queries/#mutations

### Description

I cant find any information about how to craft an update; how does graphql identify what the target(s) of the update are?

The example code is just a create (which is still unclear to me what each of these parts is)
``` graphql
# mutation statement
mutation CreateReviewForEpisode($ep: Episode!, $review: ReviewInput!) {
  createReview(episode: $ep, review: $review) {
    stars
    commentary
  }
}
# parameters
{
  "ep": "JEDI",
  "review": {
    "stars": 5,
    "commentary": "This is a great movie!"
  }
}

# result
{
  "data": {
    "createReview": {
      "stars": 5,
      "commentary": "This is a great movie!"
    }
  }
}
```
What exactly happened here? 
- `$ep : JEDI` was specified but _how was it used_?  
- What does the record look like before and after this operation? 
- What information on this mutation informs if this is 
   - added as a new record
   - appends data to an existing record
   - replaces an existing record in part or whole? 
- In the case of an update, how is the record(s) to receive the update identified in this example?

Its also missing any mention of an upsert (if exists then update, else insert). Is that missing because its just not possible?

### Motivation

My team would like to use GraphQL as an intermediary language to several backend databases instead of having to know the QL for each specific backend. I am trying to write the POC for this "translator" and I think I have a handle on data retrieval, but the available materials for mutations are either incomplete or create more questions.

### Collaboration

I'm too much of a noob to be useful in this regard.

### Additional Context

This section also makes the statement 

> While query fields are executed in parallel, mutation fields run in series, one after the other.

Which sounds like it would be an implementation dependent thing, and very inefficient. I interpret that as having some kind of business document that has Author First Name, Middle Initial, and Last Name as three fields, and if I need to change the author to someone else, the system is expected to update the First Name, then the Middle Initial, then the Last Name. To me that sounds kind of awful as it could update only 1 or 2 of the fields but not all of them.  Is my interpretation correct?

StingyJack:
I'm going to leave this open as someone may find value in it or any answer that can be provided, but we are stopping our POC of using GraphQL because even though by name it is a query language, the spec and everything after the first tutorial shows that it is at best an incomplete query language.

saihaj:
Looks like part of work @mandiwise did this addresses it. But if you think it is still missing we can re-open.

Issue #754, opened 2019-08-15
Docs should explain "data" object in results


JVimes:
Maybe I just missed it, but the docs mysteriously show a "data" object that wasn't queried for, and I can't find the explanation.

![image](https://user-images.githubusercontent.com/6148201/63122558-4a8e8600-bf64-11e9-816c-94b5dbb8579d.png)

Newbies like me could use:
* An explanation early in the docs.
* Is this a convention or requirement?
* It might be worth also mentioning the "errors" object I see in later examples.

carolstran:
👋🏼 @JVimes, thanks for opening an issue on the GraphQL documentation! I totally agree with you, this should be more clear early-on in the documentation and learning journey through the docs.

I'd lean towards adding it on the [Queries and Mutations](https://graphql.org/learn/queries/), right after the introduction and where this `data` object is first introduced. Would you agree?

I also agree with adding a similar note about the `errors` object. I wonder if that would be best when the first example is introduced (in [Schemas and Types](https://graphql.org/learn/schema/)) or mentioned in the note about `data` (or both). Would be curious to hear what you think!

JVimes:
Thank you @carolstran! Yes, I think it'll work well to define each (`data` and `errors`) with its first appearance in the docs. Consider briefly mentioning `errors` when you introduce `data`. This will satisfy curiosity as to why the data is not at the root level (to leave a place for errors).

Issue #728, opened 2019-06-26
Is the source code for the example used in /learn available? 


hacker-DOM:
I feel like the `/learn` section is a bit all over the place. It starts with Queries and Mutations, but there are more links from this chapter to Schemas and Types than vice versa. Maybe swap the order of the two chapters? In any case, I'd love a canonical source code for the tutorial - is that available? Cheers

dteiml:
Went here to look if anyone has asked this q and turns out I have a few weeks back. It's still relevant: Where can I find the source code for the tutorial? Cheers

skedwards88:
I thought the interactive code blocks were pretty nice, so I poked around to learn some more. The `MiniGraphiQL` component [here](https://github.com/graphql/graphql.github.io/blob/739a1c913fa4a88ac7c64c93ce68fc64ec0cbd96/site/_core/MiniGraphiQL.js) may be what you are looking for @hacker-DOM .

ericz1803:
I was looking for the same thing and I found the file containing the schema and resolvers [here](https://github.com/graphql/graphql.github.io/blob/739a1c913fa4a88ac7c64c93ce68fc64ec0cbd96/site/_core/swapiSchema.js).

carolstran:
👋🏼 @hacker-DOM, thanks for opening an issue on the GraphQL documentation! From what I can tell, the links that the others provided address what you're asking about - so I'm going to close this issue. If I misunderstood or something is still unclear, please let me know.

Issue #584, opened 2018-10-02
Add Subscriptions to Learn section


leebyron:
See: https://github.com/facebook/graphql/issues/334

https://graphql.org/learn/ covers Queries and Mutations but not Subscriptions.

Great task for someone who wants to contribute to the GraphQL community!

ParthS007:
Hi @leebyron 
I want to work on this Issue. How should I proceed with this?
Thanks

vnenkpet:
I'm actually surprised this still isn't covered! Seems like a huge oversight, they are actually mentioned in one place but never explained?

@ParthS007 I guess you can simply follow the contribution guide in the README and create PR for it if you feel like you can take care of this...?

tanaypratap:
@leebyron I can take this up. If no one is taking it. Have worked on subscription a lot recently and would love to contribute.

amirulazim048:
thank

carolstran:
👋🏼 👋🏼 @ParthS007 @vnenkpet @tanaypratap Would any of you still be open to taking this on? It's still relevant and I believe it'd be an important addition to the docs. I'd be happy to help structure the page, look over the copy, etc. 😊  Or you can open a PR with your proposal, reference this issue and we can take it from there!

tanaypratap:
@carolstran the GraphQL working committee has been talking about the doc site update for a long time now. 
I was wondering if there's progress in that field. We can club this and other doc related work along with it.

carolstran:
@tanaypratap Ah cool, I had no idea! I'm just trying to triage some issues while I work on research for the FAQ page 😄 Is there a good way to see/track what the working committee is discussing? Or is that secret? I'm still trying to figure out the general org structure/setup 🤪

tanaypratap:
@carolstran you can check it out here
https://github.com/graphql/graphql-wg

I haven't been able to attend the meetings for quite sometime. Last I joined, everyone, including Lee, agreed that we need an overhaul of the website. 

If you run a Lighthouse test for the current website you will see the color contrast and few other things are broken too, bringing down the a11y score overall.

PR #1802, opened 2024-10-26
Update operation docs

mandiwise:
Addresses #250, #584, #728, #754, and #992

## Description

This PR contains updated operations-related content in the Learn docs. Key changes include:

- Expanded content on GraphQL query operations
- A separate "Mutations" page with additional examples for updating and deleting data
- A new "Subscriptions" page
- Updates to the Star Wars schema to support the new mutation examples
- Operation-related content now comes after the Schemas and Types page in the left nav
- Standardization of capitalization and naming for named types

@benjie @jorydotcom


Urigo (approved):
so excited for this!

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Purpose-built mutation fields can help make a schema more expressive by allowing the input types for field arguments to be Non-Null types (a generic `updateHuman` mutation would likely need to accept many nullable arguments to handle different update scenarios). Defining this requirement in the schema also eliminates the need for other runtime logic to determine that the appropriate values were submitted to perform the client's desired write operation.
+
+Schemas should be designed to help clients get the data that they need from the GraphQL API, so the fields defined in a schema should be informed by those use cases.
+
+## Deleting data
+
+Just as we can send a `DELETE` request to delete a resource with a REST API, we can use mutations to delete some existing data as well by defining another field on the `Mutation` type:
```
saihaj, line 116: I think we should clarify that in GraphQL we send a `POST` request to delete

src/pages/learn/subscriptions.mdx
saihaj, line 1: For subscriptions I think we should mention that the community maintained specs [SSE](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md) or [WS](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) which is what most clients & server implementation are build on

src/pages/learn/queries.mdx
saihaj, line 1: since we are talking servers, I wonder if we should also mention that there is a spec for GraphQL over HTTP or just link to this doc https://graphql.org/learn/serving-over-http/ as part of next steps for production grade graphql

src/pages/learn/mutations.mdx
mandiwise: I have an updated "Serving over HTTP" page almost ready to go where I go into this in detail based on the draft GraphQL over HTTP spec (including when `GET` may be used too). Does it make sense to duplicate that here? My goal was draw connections with REST patterns the reader would likely be familiar with while explaining how different kinds of mutations work without overloading with too much information. It can be tricky to find the right balance though! Let me know what you think.

src/pages/learn/queries.mdx
mandiwise: Added a link to the recap section.

src/pages/learn/subscriptions.mdx
mandiwise: Good call out. Added links to these now.

src/pages/learn/queries.mdx
```diff
@@ -1,10 +1,22 @@
...
-At its simplest, GraphQL is about asking for specific fields on objects. Let's start by looking at a very simple query and the result we get when we run it:
+At its simplest, GraphQL is about asking for specific [fields](https://spec.graphql.org/draft/#sec-Language.Fields) on objects. Let's start by looking at a simple query for a `hero` field that's defined on the `Query` type in the schema:
+
+```graphql
+type Query {
+  hero: Character
+}
+```
+
+We can see what the result we get when we run it:
```
benjie:
````suggestion
At its simplest, GraphQL is about asking for specific [fields](https://spec.graphql.org/draft/#sec-Language.Fields) on objects. Let's start by looking at the `hero` field that's defined on the `Query` type in the schema:

```graphql
type Query {
  hero: Character
}
```

We can see what result we get when we query it:
````

File: src/pages/learn/queries.mdx [hunk 3]
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
 }
 ```
 
-You can see immediately that the query has exactly the same shape as the result. This is essential to GraphQL, because you always get back what you expect, and the server knows exactly what fields the client is asking for.
-
-The field `name` returns a `String` type, in this case the name of the main hero of Star Wars, `"R2-D2"`.
+When creating a GraphQL _document_ we always start with a [root operation type](/learn/schema/#the-query-mutation-and-subscription-types) (the `Query` Object type for this example) because it serves as an entry point to the API. From there we must specify the _selection set_ of fields all the way down to their leaf values, which will be Scalar or Enum types. The field `name` returns a `String` type, in this case the name of the main hero of Star Wars, `"R2-D2"`.
```
benjie:
```suggestion
When creating a GraphQL _document_ we always start with a [root operation type](/learn/schema/#the-query-mutation-and-subscription-types) (the `Query` Object type for this example) because it serves as an entry point to the API. From there we must specify the _selection set_ of fields we are interested in, all the way down to their leaf values which will be Scalar or Enum types. The field `name` returns a `String` type, in this case the name of the main hero of Star Wars, `"R2-D2"`.
```

src/pages/learn/queries.mdx
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
...
 
-> Oh, one more thing - the query above is _interactive_. That means you can change it as you like and see the new result. Try adding an `appearsIn` field to the `hero` object in the query, and see the new result.
+The GraphQL specification indicates that a request's result will be returned on a top-level `data` key in the response. If the request raised any errors, there will be information about what went wrong on a top-level `errors` key. From there, you can see that the result has the same shape as the query. This is essential to GraphQL, because you always get back what you expect, and the server knows exactly what fields the client is asking for.
```
benjie, line 32: Would be nice to have something like: "We'll see more on why this is so important in [link to docs]" because it's not just specifically to gain the performance and efficiency advantages of avoiding over-/under-fetching as many think, but also relates to versioning (including deprecation and field removal), error handling (preventing selection sets blowing up unnecessarily), symmetry (receiving exactly what you ask for), locality (each component declaring its own dependencies and being able to be edited without breaking other components' data needs), and shaping the response in a client-focussed way.

But no need to address this in this PR specifically.

src/pages/learn/queries.mdx
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
...
 
-In the previous example, we just asked for the name of our hero which returned a String, but fields can also refer to Objects. In that case, you can make a _sub-selection_ of fields for that object. GraphQL queries can traverse related objects and their fields, letting clients fetch lots of related data in one request, instead of making several roundtrips as one would need in a classic REST architecture.
+In the previous example, we just asked for the name of our hero which returned a `String`, but fields can also refer to more deeply nested Object types. In that case, you can make a _sub-selection_ of fields for that object:
```
benjie: I don't think this first edit is desired; we're talking about the field `name: String`, so "more deeply nested Object type" doesn't seem to make sense here since it's not returning an Object type (nested or otherwise).

```suggestion
In the previous example, we just asked for the name of our hero which returned a `String`, but fields can also return Object types (and lists thereof). In that case, you can make a _sub-selection_ of fields for that Object type:
```

src/pages/learn/queries.mdx
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
...
+In a system like REST, you can only pass a single set of arguments—the query parameters and URL segments in your request. But in GraphQL, every field and nested object can get its own set of arguments, making GraphQL a complete replacement for making multiple API fetches. 
 
-[Read more about the GraphQL type system here.](/learn/schema)
-
-## Aliases
-
-If you have a sharp eye, you may have noticed that, since the result object fields match the name of the field in the query but don't include arguments, you can't directly query for the same field with different arguments. That's why you need _aliases_ - they let you rename the result of a field to anything you want.
+You can even pass arguments into fields that output Scalar types, to implement data transformations once on the server, instead of on every client separately:
```
benjie:
```suggestion
You can even pass arguments into fields that output Scalar types; one use case for this would be to implement data transformations once on the server, instead of on every client separately:
```

src/pages/learn/queries.mdx
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
...
+Arguments can be of many different types. In the above example, we have used an Enum type, which represents one of a finite set of options (in this case, units of length, either `METER` or `FOOT`). GraphQL comes with a [default set of types](/learn/schema/#scalar-types), but a GraphQL server can also declare custom types, as long as they can be serialized into your transport format.
 
-## Fragments
+[Read more about the GraphQL type system here.](/learn/schema)
+
+## Operation type and name
 
-Let's say we had a relatively complicated page in our app, which lets us look at two heroes side by side, along with their friends. You can imagine that such a query could quickly get complicated, because we would need to repeat the fields at least once - one for each side of the comparison.
+In the examples above we have been using a shorthand syntax where we omit the `query` keyword before the operation's selection set. In addition to specifying the _operation type_ explicitly, we can also add a unique _operation name_, which is useful in production apps because it makes our code less ambiguous.
```
benjie:
```suggestion
In the examples above we have been using a shorthand syntax where we omit the `query` keyword before the operation's selection set. In addition to specifying the _operation type_ explicitly, we can also add a unique _operation name_, which is useful in production apps because it makes debugging and tracing easier.
```

src/pages/learn/queries.mdx
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
...
-# { "graphiql": true }
-query HeroComparison($first: Int = 3) {
-  leftComparison: hero(episode: EMPIRE) {
-    ...comparisonFields
-  }
-  rightComparison: hero(episode: JEDI) {
-    ...comparisonFields
-  }
-}
+The operation type is either `query`, `mutation`, or `subscription` and describes what type of operation you intend to do. This keyword is required unless you're using the shorthand syntax for queries, and it is always required for mutations and subscriptions. Additionally, if you wish to provide a name for your operation, then you must specify the operation type as well.
```
benjie:
```suggestion
The operation type is either `query`, `mutation`, or `subscription` and describes what type of operation you intend to do. This keyword is required unless you're using the shorthand syntax for queries (it is always required for mutations and subscriptions). Additionally, if you wish to provide a name for your operation, then you must specify the operation type as well.
```

src/pages/learn/queries.mdx
```diff
@@ -15,165 +27,115 @@ At its simplest, GraphQL is about asking for specific fields on objects. Let's s
...
-    totalCount
-    edges {
-      node {
-        name
-      }
-    }
-  }
-}
-```
+The operation name is a meaningful and explicit name for your operation. It is only required when sending multiple operations in one document, but it's encouraged because operation names are helpful for debugging and server-side logging. When something goes wrong (you see errors either in your network logs or in the logs of your GraphQL server) it is easier to identify a query in your codebase by name instead of trying to decipher the contents.
```
benjie:
```suggestion
The operation name is an explicit name that you assign to your operation; you should pick a meaningful name. It is required when sending multiple operations in one document, but even if you're only sending one operation it's encouraged because operation names are helpful for debugging and server-side logging. When something goes wrong (you see errors either in your network logs or in the logs of your GraphQL server) it is easier to identify a query in your codebase by name instead of trying to decipher the contents.
```

src/pages/learn/queries.mdx
```diff
@@ -224,68 +190,67 @@ query HeroNameAndFriends($episode: Episode = JEDI) {
 
 When default values are provided for all variables, you can call the query without passing any variables. If any variables are passed as part of the variables dictionary, they will override the defaults.
 
-## Directives
+## Fragments
 
-We discussed above how variables enable us to avoid doing manual string interpolation to construct dynamic queries. Passing variables in arguments solves a pretty big class of these problems, but we might also need a way to dynamically change the structure and shape of our queries using variables. For example, we can imagine a UI component that has a summarized and detailed view, where one includes more fields than the other.
+Let's say we have a relatively complicated page in our app, which lets us look at two heroes side by side, along with their friends. You can imagine that such a query could quickly get complicated because we would need to repeat the fields at least once —one for each side of the comparison.
```
benjie:
```suggestion
Let's say we have a relatively complicated page in our app, which lets us look at two heroes side by side, along with their friends. You can imagine that such a query could quickly get complicated because we would need to repeat the fields at least once—one for each side of the comparison.
```

src/pages/learn/queries.mdx
```diff
@@ -224,68 +190,67 @@ query HeroNameAndFriends($episode: Episode = JEDI) {
...
 }
 ```
 
-Try editing the variables above to instead pass `true` for `withFriends`, and see how the result changes.
-
-We needed to use a new feature in GraphQL called a _directive_. A directive can be attached to a field or fragment inclusion, and can affect execution of the query in any way the server desires. The core GraphQL specification includes exactly two directives, which must be supported by any spec-compliant GraphQL server implementation:
-
-- `@include(if: Boolean)` Only include this field in the result if the argument is `true`.
-- `@skip(if: Boolean)` Skip this field if the argument is `true`.
+You can see how the above query would be pretty repetitive if the fields were repeated. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine many UI components with different fragments into one initial data fetch.
```
benjie:
```suggestion
You can see how the above query would be pretty repetitive if we weren't able to use fragments. The concept of fragments is frequently used to split complicated application data requirements into smaller chunks, especially when you need to combine many UI components with different fragments into one initial data fetch.
```

src/pages/learn/queries.mdx
```diff
@@ -330,6 +297,54 @@ Given that there are some situations where you don't know what type you'll get b
 }
 ```
 
-In the above query, `search` returns a union type that can be one of three options. It would be impossible to tell apart the different types from the client without the `__typename` field.
+In the above query, `search` returns a Union type that can be one of three options. Without the `__typename` field, it would be impossible for a client to tell the different types apart.
+
+GraphQL services provide a few more meta fields, the rest of which are used to expose the [introspection](/learn/introspection) system.
```
benjie: "A few" is typically interpreted as "3 or more"; but I think other than `__typename` there's only `__schema` and `__type`. Let's just list them:

```suggestion
All field names beginning with two underscores (`__`) are reserved by GraphQL. In addition to `__typename`, GraphQL services provide the `__schema` and `__type` meta-fields which expose the [introspection](/learn/introspection) system.
```

src/pages/learn/queries.mdx
```diff
@@ -330,6 +297,54 @@ Given that there are some situations where you don't know what type you'll get b
...
+
+## Next steps
+
+To recap what we've learned about queries:
+
+- A GraphQL operation that reads data starts at the `query` root operation type and traverses the fields in the selection set down to the leaf values, which will be Scalar or Enum types
+- Fields can accept arguments that alter the output of that field
+- Operations can use the `query`, `mutation`, or `subscription` keyword to indicate their type 
+- The operation type keyword can be omitted for certain query operations only
+- Operations can be given unique names, which can make code more expressive and help with debugging
```
benjie:
```suggestion
- Operations should be given unique names, which make requests more expressive and help with debugging
```

src/pages/learn/queries.mdx
```diff
@@ -330,6 +297,54 @@ Given that there are some situations where you don't know what type you'll get b
...
+- Field aliases allow you to rename response keys, include the same field multiple times in the same query, and provide different arguments to the aliased fields
+- Variables are preceded by the `$` character and can be used to provide dynamic values to field arguments
+- A fragment is a reusable selection set of fields that can be used as needed in multiple queries
+- Executable directive can be applied to queries to change the result of a GraphQL query when it's executed on the server
+- All spec-compliant GraphQL servers include the `@include` and `@skip` built-in directives
 
-GraphQL services provide a few meta fields, the rest of which are used to expose the [Introspection](/learn/introspection) system.
+Not that we understand ins and outs of how to read data from a GraphQL server with query operations, so it's time to learn how to write data as well using [mutations](/learn/mutations/).
```
benjie:
```suggestion
Now that we understand the ins and outs of how to read data from a GraphQL server with query operations, it's time to learn how to change data and trigger side effects using [mutations](/learn/mutations/).
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,149 @@
+import { Callout } from "nextra/components"
+
+# Mutations
+
+<p className="learn-subtitle">Learn how to modify data with a GraphQL server</p>
+
+Most discussions of GraphQL focus on data fetching, but any complete data platform needs a way to modify server-side data as well.
+
+In REST, any request might cause some side-effects on the server, but by convention, it's suggested that one doesn't use `GET` requests to modify data. GraphQL is similar—technically any query could be implemented to cause a data write. However, it's useful to establish a convention that any operations that cause writes should be sent explicitly via a mutation.
```
benjie:
```suggestion
In REST, any request might cause some side-effects on the server, but by convention, it's suggested that one doesn't use `GET` requests to modify data. GraphQL is similar—technically any field resolver could be implemented to cause a data write—but the [GraphQL specification states](https://spec.graphql.org/draft/#sel-GANRNDAB6DBmMn6D):

> the resolution of fields other than top-level mutation fields must always be side effect-free and idempotent

Thus only the top-level fields in mutation operations are allowed to cause side effects; to do otherwise would make your schema GraphQL non-compliant.
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+On this page, you'll learn how to use mutation operations to create, update, and delete data using GraphQL.
```
benjie: I am very uncomfortable with this positioning; GraphQL isn't just a CRUD system and in general it's better to use semantically named mutations for specific user actions rather than trying to fit every mutation into the CRUD box. Some mutations such as `processPayment`, `resetPassword`, `sendEmailVerification`, `inviteUser`, `markAllNotificationsAsRead`, `redeemGiftCard` and so on (arguably) don't even fit into the CRUD box. (This isn't to say that CRUD should never be used, but more that a semantically named action should be exposed preferentially where it makes sense to do so.)

From GraphQL's point of view there are no different types of mutation - there is no difference between a create, update, delete or arbitrary user action mutation - so I'm concerned that highlighting CRUD mutations in this way sets the reader (especially someone who does not read the page fully) up for misconceptions.

benjie (changes requested):
I love the improvements to the queries page (and include a number of minor suggested edits), but I have concerns over the mutations page. I have not yet read the subscriptions page.

src/pages/learn/mutations.mdx
mandiwise: I see where you're coming from. I think the challenge here is to figure out how much information to include up-front in the intro for a reader who has some experience with REST and needs to connect those dots with GraphQL (e.g. #992). This distinction is called out in the paragraph that follows the new example for updating a human name, but it's halfway down the page.

Do you think it would be better to generalize it like this?

> On this page, you'll learn how to use mutation operations to modify data on a server using GraphQL.

mandiwise: Do you think the subheads on the page are problematic? We could further deemphasize any kind of 1:1 CRUD implication by renaming them like this:

~Create data~ Add new data
~Update data~ Update existing data
~Delete data~ Remove existing data

mandiwise: One more thought on this is—I think there's a gap in the Best Practices content where "migrating from REST" could be covered in more detail too with a few more detailed examples that are based on a further-enhanced version of the Star Wars schema.

benjie: I'm not sure that our learn docs should necessarily cater primarily to people who already know REST/CRUD - perhaps a separate guide for "how to convert CRUD REST operations to GraphQL" might make sense but I think our learn docs shouldn't start with REST concepts and map them to GraphQL, but instead start with GraphQL best practices and then if necessary expand with REST accommodations/analogies.

I think my preferred way of addressing this would be to start with a non-CRUD mutation ~~that follows best practices (input object, mutation payload)~~ so that that is front and centre, and then maybe have a sub heading about CRUD that has examples of create, update, upsert, delete. I think the original documentation used a create mutation because that's fairly universal, but perhaps pivoting away from that would help learners to do GraphQL the right way from the start?

Trying to think of something within the Star Wars schema that would make sense as a non-CRUD mutation, perhaps `rateFilm` which stores a rating "onto the film" from your perspective, but clearly isn't actually updating the film itself:

```graphql
type Mutation {
  rateFilm(input: RateFilmInput!): RateFilmPayload
}

input RateFilmInput {
  filmId: ID!
  rating: FilmRating! = THUMBS_UP
}

type RateFilmPayload {
  film: Film
}

type Film {
  # ...
  viewerRating: FilmRating
}

enum FilmRating {
  THUMBS_DOWN
  NEUTRAL
  THUMBS_UP
  DOUBLE_THUMBS_UP
}
```

benjie: Actually I think this page should talk about the input object/payload best practices at the end, so maybe simplify this down:

```graphql
type Mutation {
  rateFilm(
    filmId: ID!
    rating: FilmRating! = THUMBS_UP
  ): Film
}

type Film {
  # ...
  viewerRating: FilmRating
}

enum FilmRating {
  THUMBS_DOWN
  NEUTRAL
  THUMBS_UP
  DOUBLE_THUMBS_UP
}
```

mandiwise: I can come up with something but at this point I think I'll wait until #1812 is merged so I rebase the changes made in `swapi-schema.tsx` in that PR first.

benjie: - #1812

is now merged :+1:

mandiwise: There's a new commit with the requested changes included now.

benjie:
Not quite sure what happened here but it went to 136 files changed... I've merged in the latest `source` and resolved the conflicts... That seems to have us back down to 5 files changed.

src/pages/learn/subscriptions.mdx
```diff
@@ -0,0 +1,102 @@
+# Subscriptions
+
+import { Callout } from "nextra/components"
+
+<p className="learn-subtitle">Learn how to get real-time updates from a GraphQL server</p>
+
+In addition to reading and writing data using stateless query and mutation operations, the GraphQL specification also describes how to fetch data in real-time via long-lived requests. On this page, we'll explore how clients can receive updated data in response to some event on a GraphQL server using subscription operations.
```
benjie: Edited for precision (events don't necessarily have to update existing data; fetch implies client-pull but subscriptions are server-push):

```suggestion
In addition to reading and writing data using stateless query and mutation operations, the GraphQL specification also describes how to receive real-time updates via long-lived requests. On this page, we'll explore how clients can subscribe to details of events on a GraphQL server using subscription operations.
```

src/pages/learn/subscriptions.mdx
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+```
+
+GraphQL subscriptions are typically backed by a separate pub/sub system so that messages about updated data can be published as needed during runtime and then consumed by the resolver functions for the subscription fields in the API.
+
+For the previous example, we could imagine publishing a `REVIEW_CREATED` message to a pub/sub system when a client submits a new review with the `createReview` mutation, and then listening for this event in the resolver function for the `reviewCreated` subscription field so that the server may send this data to subscribed clients.
+
+<Callout type="info">
+Note that subscriptions are different from the concept of _live queries_, which are not formally described in the specification and are often implemented differently in GraphQL various implementations using an `@live` executable directive. [Read more about the ongoing discussion on a formal specification for live queries here.](https://github.com/graphql/graphql-spec/issues/386)
```
benjie: Let's make it a little clearer that there's no standard implementation of live queries.

```suggestion
Note that subscriptions are different from the concept of _live queries_, a loosely defined feature that some GraphQL implementations may offer. [Read more about the ongoing discussion on a formal specification for live queries here.](https://github.com/graphql/graphql-spec/issues/386)
```

src/pages/learn/subscriptions.mdx
```diff
@@ -0,0 +1,102 @@
...
+  humanFriendsUpdated {
+    name
+    friends {
+      name
+    }
+  }
+}
+```
+
+But this operation would run:
```
benjie:
```suggestion
But this document is valid:
```

src/pages/learn/subscriptions.mdx
```diff
@@ -0,0 +1,102 @@
...
+  humanFriendsUpdated(id: $id) {
+    name
+    friends {
+      name
+    }
+  }
+}
+```
+
+As with query and mutation operations, the document above contains multiple subscription operations so each operation must be named.
```
benjie:
```suggestion
As with query and mutation operations, the document above contains multiple operations so each operation must be named, and the name of the operation to execute must be specified in the request.
```

src/pages/learn/subscriptions.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+Similarly, client libraries will typically need more advanced features to handle these kinds of operations, such as logic to resubscribe to operations if the connection is disrupted or mechanisms to handle potential race conditions between initial queries and updated subscription data.
+
+In practice, a GraphQL API that supports subscriptions will call for a more complicated architecture than one that only exposes query and mutation fields. How this architecture is designed will depend on the specific GraphQL implementation, what pub/sub system supports the subscriptions, and the chosen transport protocol, as well as other requirements related to availability, scalability, and security of the real-time data.
+
+Subscription operations are best suited for data that changes often and incrementally, and for clients that need to receive those incremental updates as close to real-time as possible to deliver the expected user experience. For less frequent updates, periodic polling, mobile push notifications, or re-fetching queries based on user interaction may be the better solutions for keeping a client UI up to date.
```
benjie: The text talked about subscriptions relating to "data that changes", but subscriptions are really well suited for events such as "Bob is typing..." or "Updated GraphQL schema available" that may not actually have a representation in the data and thus polling may not even be an option. I've tweaked this wording so that it doesn't exclude the pure-event use case:

```suggestion
Subscription operations are well suited for data that changes often and incrementally, and for clients that need to receive those incremental updates as close to real-time as possible to deliver the expected user experience. For data with less frequent updates, periodic polling, mobile push notifications, or re-fetching queries based on user interaction may be the better solutions for keeping a client UI up to date.
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,161 @@
...
+}
+
+type Mutation {
+  createReview(episode: Episode, review: ReviewInput!): Review
+}
+```
+
+Like queries, mutation fields are added to one of the [root operation types](https://spec.graphql.org/draft/#sec-Root-Operation-Types) that provide an entry point to the API. In this case, we define the `createReview` field on the `Mutation` type. 
+
+Mutation fields can also accept arguments and you might notice that the `review` argument has an input type set to `ReviewInput`. This is known as [Input Object type](/learn/schema/#input-object-types), which allows us to pass in a whole object to be created instead of individual scalar values only.
```
benjie: Let's not imply that the input object itself is "created":

```suggestion
Mutation fields can also accept arguments and you might notice that the `review` argument has an input type set to `ReviewInput`. This is known as [Input Object type](/learn/schema/#input-object-types), which allows us to pass in a structured object containing information the mutation can use instead of individual scalar values only.
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,161 @@
...
+# { "graphiql": true, "variables": { "ep": "JEDI", "review": { "stars": 5, "commentary": "This is a great movie!" } } }
+mutation CreateReviewForEpisode($ep: Episode!, $review: ReviewInput!) {
+  createReview(episode: $ep, review: $review) {
+    stars
+    commentary
+  }
+}
+```
+
+While the `createReview` field could be defined with any valid output type in the schema, it's conventional to specify an output type that corresponds to whatever is modified during the mutation—in this case, the `Review` type. This can be useful for clients that need to fetch the new state of an object after an update.
```
benjie: I don't want to imply that `createReview(...): Review` is the convention people should follow when `createReview(...): CreateReviewPayload` is a much better pattern. I've changed "corresponds" to "relates" since the `CreateReviewPayload` and `Review` both _relate_ to the type that's modified, whereas I'm not sure they both "correspond" to it? (I'm no English scholar though!)

```suggestion
While the `createReview` field could be defined with any valid output type in the schema, it's conventional to specify an output type that relates to whatever is modified during the mutation—in this case, the `Review` type. This can be useful for clients that need to fetch the new state of an object after an update.
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,161 @@
...
+Mutation: {
+  createReview(obj, args, context, info) {
+    return context.db
+      .createNewReview(args.episode, args.review)
+      .then((reviewData) => new Review(reviewData))
+  }
+}
+```
+
+You can learn more about GraphQL provides data for fields on the [Execution page](/learn/execution).
```
benjie: Missing word?

```suggestion
You can learn more about how GraphQL provides data for fields on the [Execution page](/learn/execution).
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,161 @@
...
+# { "graphiql": true, "variables": { "id": "1000", "name": "Luke Starkiller" } }
+mutation UpdateHumanName($id: ID!, $name: String!) { 
+  updateHumanName(id: $id, name: $name ) {
+    id
+    name
+  }
+}
+```
+
+This example demonstrates an important distinction from REST. To update a human's properties using a REST API, you would likely send any updated data to a generalized endpoint for that resource using a `PATCH` request. With GraphQL, instead of simply creating an `updateHuman` mutation, you can define more specific mutation fields such as `updateHumanName`that are designed for the task at hand.
```
benjie: This is already really good, but I've edited it to make it even clearer this is a choice.

```suggestion
This example demonstrates an important distinction from REST. To update a human's properties using a REST API, you would likely send any updated data to a generalized endpoint for that resource using a `PATCH` request. With GraphQL, instead of creating a generic `updateHuman` mutation, you may choose to define more specific mutation fields such as `updateHumanName` that are designed for the task at hand.
```

src/pages/learn/mutations.mdx
```diff
@@ -0,0 +1,161 @@
...
+# { "graphiql": true, "variables": { "episode": "EMPIRE", "rating": "THUMBS_UP" } }
+mutation RateFilm($episode: Episode!, $rating: FilmRating!) { 
+  rateFilm(episode: $episode, rating: $rating) {
+    episode
+    viewerRating
+  }
+}
+```
+
+As a general rule, schemas should be designed to help clients get the data that they need from the GraphQL API, so the fields defined in a schema should be informed by those use cases.
```
benjie: This is really great content, but "get the data that they need from" seems very query-focussed (and side-effect free) - can you tweak it so that it better reflects mutations?

src/pages/learn/mutations.mdx
benjie: Should this section have it's own header: `## Purpose-driven mutations`

benjie (changes requested):
I like this new structure and this is great content - really really good! I've included some minor nit-picky suggestions below - the only big change I think is that it would be nice to see the purpose-driven mutations content raised to be its own heading so it's easy to link to, and doing so may or may not require some minor reworking of the surrounding text.

src/pages/learn/mutations.mdx
mandiwise: Fixed in https://github.com/graphql/graphql.github.io/pull/1802/commits/1f3ae06e30c231b965c8f591ababce6d9c88ae8f.

benjie: Perfect 🙌

benjie (approved):
Looks great to me!

PR #1803, opened 2024-10-30
Update validation page.

mandiwise:
## Description

This PR contains updated validation content for the Learn docs. Key changes include:

- New section that discusses how/when validation errors are returned to a client (note: I have a new "Response" page in development, and I will add a link on the validation page that points to the new page for more info when it's ready)
- Subheads for validation error examples to improve readability
- Light editing throughout

@benjie @jorydotcom

### DIFF
--- a/src/pages/learn/validation.mdx
+++ b/src/pages/learn/validation.mdx
@@ -1,22 +1,107 @@
 # Validation
 
-By using the type system, it can be predetermined whether a GraphQL query
-is valid or not. This allows servers and clients to effectively inform
-developers when an invalid query has been created, without having to rely
-on runtime checks.
+<p className="learn-subtitle">Learn how GraphQL validates operations using a schema</p>
 
-For our Star Wars example, the file
-[starWarsValidation-test.ts](https://github.com/graphql/graphql-js/blob/main/src/__tests__/starWarsValidation-test.ts)
-contains a number of queries demonstrating various invalidities, and is a test
-file that can be run to exercise the reference implementation's validator.
+On this page, we'll explore an important phase in the lifecycle of a GraphQL request called [validation](https://spec.graphql.org/draft/#sec-Validation). A request must be syntactically correct to run, but it should also be valid when checked against the API's schema. 
 
-To start, let's take a complex valid query. This is a nested query, similar to
-an example from the previous section, but with the duplicated fields factored
-out into a fragment:
+In practice, when a GraphQL operation reaches the server, the document is first parsed and then validated using the type system. This allows servers and clients to effectively inform developers when an invalid query has been created, and without relying on runtime checks. Once the operation is validated, it can be [executed](/learn/execution/) on the server and a response will be delivered to the client.
+
+## Validation examples
+
+The GraphQL specification describes the detailed conditions that must be satisfied for a request to be considered valid. In the sections that follow, we'll look at a few examples of common validation issues that occur in GraphQL operations.
+
+### Requesting non-existent fields
+
+When we query for fields, we have to request a field on the given type. So as `hero` returns a `Character` type, we have to query for a field that's defined on `Character`. That type does not have a `favoriteSpaceship` field, so this query is invalid:
 
 ```graphql
 # { "graphiql": true }
-{
+# INVALID: favoriteSpaceship does not exist on Character
+query {
+  hero {
+    favoriteSpaceship
+  }
+}
+```
+
+### Selecting invalid leaf fields
+
+Whenever we query for a field and it returns something other than a Scalar or Enum type, we need to specify what data we want to get back from the field. The `hero` query field returns a `Character`, and we've already seen examples that request fields like `name` and `appearsIn` on it. If we omit those leaf field selections, then the query will not be valid:
+
+```graphql
+# { "graphiql": true }
+# INVALID: hero is not a scalar, so fields are needed
+query {
+  hero
+}
+```
+
+Similarly, if the leaf field is a scalar value, it doesn't make sense to query for additional fields on it, and doing so will make the query invalid:
+
+```graphql
+# { "graphiql": true }
+# INVALID: name is a scalar, so fields are not permitted
+query {
+  hero {
+    name {
+      firstCharacterOfName
+    }
+  }
+}
+```
+
+### Missing fragments for fields that output abstract types
+
+Earlier, it was noted that a query can only ask for fields on the type in question. So when we query for `hero` which returns a `Character`, we can only request fields that exist on the `Character` Interface type. What happens if we want to query for R2-D2's primary function, though?
+
+```graphql
+# { "graphiql": true }
+# INVALID: primaryFunction does not exist on Character
+query {
+  hero {
+    name
+    primaryFunction
+  }
+}
+```
+
+That query is invalid, because `primaryFunction` is not one of the shared fields defined by the `Character` Interface type. We want some way of indicating that we wish to fetch `primaryFunction` if the `Character` is a `Droid`, and to ignore that field otherwise. We can use the [fragments](/learn/queries/#fragments) to do this. By setting up a fragment defined on `Droid` and including it in the selection set, we ensure that we only query for `primaryFunction` where it is defined:
+
+```graphql
+# { "graphiql": true }
+query {
+  hero {
+    name
+    ...DroidFields
+  }
+}
+
+fragment DroidFields on Droid {
+  primaryFunction
+}
+```
+
+This query is valid, but it's a bit verbose; named fragments were valuable above when we used them multiple times, but we're only using this one once. Instead of using a named fragment, we can use an [inline fragment](/learn/queries/#inline-fragments); this still allows us to indicate the type we are querying on, but without naming a separate fragment:
+
+```graphql
+# { "graphiql": true }
+query {
+  hero {
+    name
+    ... on Droid {
+      primaryFunction
+    }
+  }
+}
+```
+
+### Cyclic fragment spreads
+
+To start, let's take a complex valid query. This is a nested query, but with the duplicated fields factored out into a fragment:
+
+```graphql
+# { "graphiql": true }
+query {
   hero {
     ...NameAndAppearances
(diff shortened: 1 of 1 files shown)


src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
-For our Star Wars example, the file
-[starWarsValidation-test.ts](https://github.com/graphql/graphql-js/blob/main/src/__tests__/starWarsValidation-test.ts)
-contains a number of queries demonstrating various invalidities, and is a test
-file that can be run to exercise the reference implementation's validator.
+On this page, we'll explore an important phase in the lifecycle of a GraphQL request called [validation](https://spec.graphql.org/draft/#sec-Validation). A request must be syntactically correct to run, but it should also be valid when checked against the API's schema. 
 
-To start, let's take a complex valid query. This is a nested query, similar to
-an example from the previous section, but with the duplicated fields factored
-out into a fragment:
+In practice, when a GraphQL operation reaches the server, the document is first parsed and then validated using the type system. This allows servers and clients to effectively inform developers when an invalid query has been created, and without relying on runtime checks. Once the operation is validated, it can be [executed](/learn/execution/) on the server and a response will be delivered to the client.
```
benjie:
```suggestion
In practice, when a GraphQL operation reaches the server, the document is first parsed and then validated using the type system. This allows servers and clients to effectively inform developers when an invalid query has been created, without relying on runtime checks. Once the operation is validated, it can be [executed](/learn/execution/) on the server and a response will be delivered to the client.
```

src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
+## Validation examples
 
-```graphql
-# { "graphiql": true }
-{
-  hero {
-    ...NameAndAppearancesAndFriends
-  }
-}
+The GraphQL specification describes the detailed conditions that must be satisfied for a request to be considered valid. In the sections that follow, we'll look at a few examples of common validation issues that occur in GraphQL operations.
```
benjie:
```suggestion
The GraphQL specification describes the detailed conditions that must be satisfied for a request to be considered valid. In the sections that follow, we'll look at a few examples of common validation issues that may occur in GraphQL operations.
```

src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
-  }
-}
-```
+### Requesting non-existent fields
 
-When we query for fields, we have to query for a field that exists on the
-given type. So as `hero` returns a `Character`, we have to query for a field
-on `Character`. That type does not have a `favoriteSpaceship` field, so this
-query is invalid:
+When we query for fields, we have to request a field on the given type. So as `hero` returns a `Character` type, we have to query for a field that's defined on `Character`. That type does not have a `favoriteSpaceship` field, so this query is invalid:
```
benjie: I recognise that this was the original text, but it doesn't read well to me. Here's a proposed alternative, what do you think?

```suggestion
When we query for a field, the field must be defined on the relevant type. As `hero` returns a `Character` type, its selection set may only request the `Character` type's fields; `Character` does not have a `favoriteSpaceship` field, so this query is invalid:
```

Incidentally it would be much nicer to write `Query.hero` here, but to do so we'd have to introduce "schema coordinates" earlier.

src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
 
 ```graphql
 # { "graphiql": true }
 # INVALID: favoriteSpaceship does not exist on Character
-{
+query {
```
benjie, line 20: Interested to know why you're moving away from the query shorthand here? It tends to be very rare that I see this - either the operation has a name and/or variables in which case you need `query`, or it has neither in which case it's very rare to see it in my experience.

src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
     favoriteSpaceship
   }
 }
 ```
 
-Whenever we query for a field and it returns something other than a scalar
-or an enum, we need to specify what data we want to get back from the field.
-Hero returns a `Character`, and we've been requesting fields like `name` and
-`appearsIn` on it; if we omit that, the query will not be valid:
+### Selecting invalid leaf fields
```
benjie: Perhaps this is a clearer title?

```suggestion
### Selection sets and leaf fields
```

src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
+
+Whenever we query for a field and it returns something other than a Scalar or Enum type, we need to specify what data we want to get back from the field. The `hero` query field returns a `Character`, and we've already seen examples that request fields like `name` and `appearsIn` on it. If we omit those leaf field selections, then the query will not be valid:
```
benjie:
```suggestion
Whenever we query for a field and it returns something other than a Scalar or Enum type, we need to specify what data we want to get back from the field (a "selection set"). The `hero` query field returns a `Character`, and we've already seen examples that request fields like `name` and `appearsIn` on it. If we omit those leaf field selections, then the query will not be valid:
```

src/pages/learn/validation.mdx
```diff
@@ -1,97 +1,47 @@
...
 # INVALID: hero is not a scalar, so fields are needed
-{
+query {
   hero
 }
 ```
 
-Similarly, if a field is a scalar, it doesn't make sense to query for
-additional fields on it, and doing so will make the query invalid:
+Similarly, if the leaf field is a scalar value, it doesn't make sense to query for additional fields on it, and doing so will make the query invalid:
```
benjie: This statement is true of leaf fields whether or not they are scalar values (also applies to enums); maybe this edit is clearer?

```suggestion
Similarly, querying fields of a scalar or enum doesn’t make sense, therefore adding a selection set to a leaf field will make the query invalid:
```

Alternatively maybe something like "Similarly, since a leaf field represents a scalar or enum, it doesn't make sense..."?

src/pages/learn/validation.mdx
```diff
@@ -155,10 +95,65 @@ a separate fragment:
...
+
+fragment NameAndAppearances on Character {
+  name
+  appearsIn
+}
+```
+
+And this query is valid. Let's take a look at some invalid queries...
+
+A fragment cannot refer to itself or create a cycle, as this could result in an unbounded result! Here's the same query above but without the explicit three levels of nesting:
```
benjie: We mentioned the query was valid already and we only have one query below here; here's an attempt to merge these paragraphs:

```suggestion
The following is an alternative to the above query, attempting to use recursion instead of the explicit three levels of nesting. This new query is invalid because a fragment cannot refer to itself (directly or indirectly) since the resulting cycle could create an unbounded result!
```

src/pages/learn/validation.mdx
```diff
@@ -155,10 +95,65 @@ a separate fragment:
...
+fragment NameAndAppearancesAndFriends on Character {
+  name
+  appearsIn
+  friends {
+    ...NameAndAppearancesAndFriends
+  }
+}
+```
+
+This has just scratched the surface of the validation system; there are a number of validation rules in place to ensure that a GraphQL operation is semantically meaningful. The specification goes into more detail about this topic in the [validation section](https://spec.graphql.org/draft/#sec-Validation), and the [validation directory](https://github.com/graphql/graphql-js/blob/main/src/validation) in GraphQL.js contains code implementing a specification-compliant GraphQL validator.
```
benjie:
```suggestion
This has just scratched the surface of the validation system; there are a number of validation rules in place to ensure that a GraphQL operation is semantically meaningful. The specification goes into more detail about this topic in the [validation section](https://spec.graphql.org/draft/#sec-Validation), and the [validation directory in the reference implementation](https://github.com/graphql/graphql-js/blob/main/src/validation) contains code implementing a specification-compliant GraphQL validator.
```

src/pages/learn/validation.mdx
```diff
@@ -155,10 +95,65 @@ a separate fragment:
...
+
+And because the GraphQL specification requires all implementations to validate incoming requests against the schema, developers won't need to write specific runtime logic to handle these validation issues manually.
+
+## Next steps
+
+To recap what we've learned about validation:
+
+- The GraphQL documents that clients submit in their requests must be syntactically correct and considered valid when checked against the schema
+- The GraphQL specification requires that implementations check that incoming requests contain valid field selections, correct fragment usage, and more
+- When a validation issue occurs, the server will raise a request error and return information about what happened to the client before field execution occurs
```
benjie:
```suggestion
- To be executed, requests must include a syntactically correct document that is considered valid when checked against the schema
- The specification requires implementations check incoming requests contain valid field selections, correct fragment usage, and more
- When a validation issue occurs, the server will raise a request error and return to the client information about what happened; field execution will not start
```

src/pages/learn/validation.mdx
```diff
@@ -155,10 +95,65 @@ a separate fragment:
...
+
+Head over to the [Execution](/learn/execution/) page to learn how GraphQL provides data for each field in a request after the validation step is complete.
```
benjie:
```suggestion
Head over to the [Execution](/learn/execution/) page to learn how GraphQL provides data for each field in a request after the validation step successfully completes.
```

benjie (changes requested):
This is looking good! Here are some light editorial suggestions (many of which apply to the original content).

src/pages/learn/validation.mdx
mandiwise: I've tended to see the opposite (inclusion of the `query` keyword) so I included it for clarity after all of the keywords are introduced on the operation pages. But I can remove it throughout if the shorthand syntax is the preferred convention.

benjie: Let's worry about that afterwards, if at all; it's much easier to replace `/^query {$/` with `{` than the reverse - so keep going with the `query` keyword for now.

Interesting that your experience differs; I wonder what the ecosystem as a whole sees!

benjie (approved):
Looks good to me! :raised_hands:

PR #1804, opened 2024-10-30
Update execution docs

mandiwise:
## Description

This PR contains updated execution content for the Learn docs. Key changes include:

- Light editing and better linking throughout the Execution page
- The format for resolver examples is standardized to other examples in the previous page updates
- A new "Response" page that describes the `data` and `errors` keys in detail, including examples for different syntax, validation, and field error scenarios

@benjie @jorydotcom

### DIFF
--- a/src/pages/learn/execution.mdx
+++ b/src/pages/learn/execution.mdx
@@ -1,8 +1,14 @@
+import { Callout } from "nextra/components"
+
 # Execution
 
-After being validated, a GraphQL query is executed by a GraphQL server which returns a result that mirrors the shape of the requested query, typically as JSON.
+<p className="learn-subtitle">Learn how GraphQL provides data for requested fields</p>
 
-GraphQL cannot execute a query without a type system, let's use an example type system to illustrate executing a query. This is a part of the same type system used throughout the examples in these articles:
+After a parsed document is [validated](/learn/validation/), a client's request will be [executed](https://spec.graphql.org/draft/#sec-Execution) by the GraphQL server and the returned result will mirror the shape of the requested query. On this page, you'll learn about the execution phase of GraphQL operations where data is read from or written to an existing source depending on what fields are requested by a client.
+
+## Field resolvers
+
+GraphQL cannot execute operations without a type system, so let's use an example type system to illustrate executing a query. This is a part of the same type system used throughout the examples in this guide:
 
 ```graphql
 type Query {
@@ -26,11 +32,11 @@ type Starship {
 }
 ```
 
-In order to describe what happens when a query is executed, let's use an example to walk through.
+To describe what happens when a query is executed, let's walk through an example:
 
 ```graphql
 # { "graphiql": true }
-{
+query {
   human(id: 1002) {
     name
     appearsIn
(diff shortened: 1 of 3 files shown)


src/pages/learn/execution.mdx
```diff
@@ -41,66 +47,66 @@ In order to describe what happens when a query is executed, let's use an example
...
+If a field produces a scalar value like a string or number, then the execution completes. However, if a field produces an object value then the query will contain another selection of fields that apply to that object. This continues until the leaf values are reached. GraphQL queries always end at Scalar or Enum types.
 
-## Root fields & resolvers
+## Root fields and resolvers
 
-At the top level of every GraphQL server is a type that represents all of the possible entry points into the GraphQL API, it's often called the _Root_ type or the _Query_ type.
+At the top level of every GraphQL server is an Object type that represents the possible entry points into the GraphQL API, it's often called the `query` root operation type or the `Query` type. If the API supports mutations to write data and subscriptions to fetch real-time data as well, then it will have `Mutation` and `Subscription` types that expose fields to perform these kinds of operations too. You can learn more about these types on the [Schema and Types page](/learn/schema/#the-query-mutation-and-subscription-types).
```
benjie: I felt like having `query` and `Query` here might be confusing, perhaps adding quotes to clarify that the whole phrase "the query root operation type" is what it's often called would clarify this. Feel free to disregard if you don't agree.

```suggestion
At the top level of every GraphQL server is an Object type that represents the possible entry points into the GraphQL API, it's often called "the `query` root operation type" or the `Query` type. If the API supports mutations to write data and subscriptions to fetch real-time data as well, then it will have `Mutation` and `Subscription` types that expose fields to perform these kinds of operations too. You can learn more about these types on the [Schema and Types page](/learn/schema/#the-query-mutation-and-subscription-types).
```

src/pages/learn/execution.mdx
```diff
@@ -41,66 +47,66 @@ In order to describe what happens when a query is executed, let's use an example
...
-  }
+function Query_human(obj, args, context, info) {
+  return context.db.loadHumanByID(args.id).then(
+    userData => new Human(userData)
+  )
 }
 ```
 
-This example is written in JavaScript, however GraphQL servers can be built in [many different languages](/code/). A resolver function receives four arguments:
+This example is written in JavaScript, however, GraphQL servers can be built in [many different languages](/code/). The resolver function receives four arguments:
```
benjie: `resolveInfo` is not required (or even mentioned) by the spec; neither is `context` actually. So let's clarify that this is specifically how the reference implementation works, and that this is how it works for all reference implementation resolvers:

```suggestion
This example is written in JavaScript, however GraphQL servers can be built in [many different languages](/code/). In the reference implementation, a resolver function receives four arguments:
```

src/pages/learn/execution.mdx
```diff
@@ -41,66 +47,66 @@ In order to describe what happens when a query is executed, let's use an example
...
 
-- `obj` The previous object, which for a field on the root Query type is often not used.
-- `args` The arguments provided to the field in the GraphQL query.
-- `context` A value which is provided to every resolver and holds important contextual information like the currently logged in user, or access to a database.
-- `info` A value which holds field-specific information relevant to the current query as well as the schema details, also refer to [type GraphQLResolveInfo for more details](/graphql-js/type/#graphqlobjecttype).
+- `obj`: The previous object (for a field on the root `Query` type, this argument is often not used).
+- `args`: The arguments provided to the field in the GraphQL operation.
+- `context`: A value provided to every resolver that may hold important contextual information like the currently logged in user, or access to a database.
+- `info`: A value holding field-specific information relevant to the current operation as well as the schema details. Refer to [type GraphQLResolveInfo for more details](/graphql-js/type/#graphqlobjecttype).
```
benjie: We should try and make it clear that `info` isn't something most people need to concern themselves with - at least not at first!

```suggestion
- `info`: generally only used in advanced use-cases, this is a value holding field-specific information relevant to the current operation as well as the schema details; refer to [type GraphQLResolveInfo](/graphql-js/type/#graphqlobjecttype) for more details.
```

src/pages/learn/execution.mdx
```diff
@@ -41,66 +47,66 @@ In order to describe what happens when a query is executed, let's use an example
...
     userData => new Human(userData)
   )
 }
 ```
 
-The `context` is used to provide access to a database which is used to load the data for a user by the `id` provided as an argument in the GraphQL query. Since loading from a database is an asynchronous operation, this returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). In JavaScript, Promises are used to work with asynchronous values, but the same concept exists in many languages, often called _Futures_, _Tasks_ or _Deferred_. When the database returns, we can construct and return a new `Human` object.
+The `context` is used to provide access to a database. Specifically, the `id` argument in the GraphQL query is used to fetch data for a user. Since loading from a database is an asynchronous operation, this returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). In JavaScript, Promises are used to work with asynchronous values, but the same concept exists in many languages, often called _Futures_, _Tasks_, or _Deferred_. When the database returns the data, we can construct and return a new `Human` object.
```
benjie: I don't think this "specifically" fits here. Here's an alternative take:

```suggestion
The `id` argument in the GraphQL query specifies the user whose data is requested, while `context` provides access to retrieve this data from a database. Since loading from a database is an asynchronous operation, this returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). In JavaScript, Promises are used to work with asynchronous values, but the same concept exists in many languages, often called _Futures_, _Tasks_, or _Deferred_. When the database returns the data, we can construct and return a new `Human` object.
```

Also... It isn't 2015 any more, `async/await` is widespread now. It would be nice to rewrite the function (maybe in a follow-up PR?) so it's easier to follow and we don't have to talk so much about promises:

```ts
async function Query_human(obj, args, context, info) {
  const userData = await context.db.loadHumanByID(args.id);
  return new Human(userData);
}
```

src/pages/learn/execution.mdx
```diff
@@ -41,66 +47,66 @@ In order to describe what happens when a query is executed, let's use an example
...
-  name(obj, args, context, info) {
-    return obj.name
-  }
+function Human_name(obj, args, context, info) {
+  return obj.name
 }
 ```
 
-A GraphQL server is powered by a type system which is used to determine what to do next. Even before the `human` field returns anything, GraphQL knows that the next step will be to resolve fields on the `Human` type since the type system tells it that the `human` field will return a `Human`.
+A GraphQL server is powered by a type system that is used to determine what to do next. Even before the `human` field returns anything, GraphQL knows that the next step will be to resolve fields on the `Human` type since the type system tells it that the `human` field will return this output type.
```
benjie:
```suggestion
A GraphQL server is powered by a type system that is used to determine what to do next. Even before the `human` field returns anything, GraphQL knows the next step will be to resolve fields on the `Human` type since the type system tells it that the `human` field will return this output type.
```

src/pages/learn/execution.mdx
```diff
@@ -41,66 +47,66 @@ In order to describe what happens when a query is executed, let's use an example
...
 
-Resolving the name in this case is very straight-forward. The name resolver function is called and the `obj` argument is the `new Human` object returned from the previous field. In this case, we expect that Human object to have a `name` property which we can read and return directly.
+Resolving the name in this case is straightforward. The name resolver function is called and the `obj` argument is the new `Human` object returned from the previous field. In this case, we expect this object to have a `name` property, which we can read and return directly.
 
-In fact, many GraphQL libraries will let you omit resolvers this simple and will just assume that if a resolver isn't provided for a field, that a property of the same name should be read and returned.
+Many GraphQL libraries will let you omit resolvers this simple and will assume that if a resolver isn't provided for a field, a property of the same name should be read and returned.
```
benjie:
```suggestion
Many GraphQL libraries let you omit resolvers this simple, assuming that if a resolver isn't provided for a field, a property of the same name should be read and returned.
```

src/pages/learn/execution.mdx
```diff
@@ -156,3 +160,18 @@ Let's take one last look at the original query to see how all these resolving fu
...
 ```
+
+As we can see, each field in the nested selection sets resolves to a scalar leaf value during execution of the query.
+
+## Next steps
+
+To recap what we've learned about execution:
+
+- Each field in a GraphQL type system will have a corresponding resolver function that provides data for the field from an existing data source
+- Execution begins at the top-level `query`, `mutation`, and `subscription` fields
```
benjie: Should this be:

```suggestion
- Execution begins at the top-level `Query`, `Mutation`, or `Subscription` fields
```

It's not clear if the fields on the type are being referenced, or the top-level selections in the operation.

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+}
+```
+
+The inclusion of the `data` key isn't an arbitrary decision made by the underlying GraphQL implementation—it's described by the [GraphQL specification](https://spec.graphql.org/draft/#sec-Data). Under this key, you will find the result of the execution of the requested operation, which may include partial data for the requested fields if errors were raised during the execution of some field resolvers.
+
+One thing that the GraphQL specification doesn't require is a specific serialization format for the response. That said, responses are typically formatted as JSON, as in the example above.
+
+Additionally, the GraphQL specification doesn't require the use of a particular transport protocol for requests either, although it is [common for HTTP to be used](/learn/serving-over-http/) with `POST` or `GET` methods for stateless query and mutations operations. Long-lived, stateful subscription operations are often supported by WebSockets or server-sent events instead.
```
benjie: I understand the intent, but this could be read to imply `POST` is used for query and `GET` for mutations. Without expanding the text, just re-ordering the methods would partially mitigate the effect of incorrect interpretation. Or maybe just drop "with `GET` or `POST` methods" entirely:

```suggestion
Additionally, the GraphQL specification doesn't require the use of a particular transport protocol for requests either, although it is [common for HTTP to be used](/learn/serving-over-http/) for stateless query and mutations operations. Long-lived, stateful subscription operations are often supported by WebSockets or server-sent events instead.
```

Alternatively feel free to use more precise wording or suggest another fix.

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+<Callout type="info">
+There is [a draft GraphQL over HTTP specification](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md) available with further guidelines for using HTTP with a GraphQL clients and servers.
```
benjie:
```suggestion
There is [a draft GraphQL over HTTP specification](https://graphql.github.io/graphql-over-http/draft/) available with further guidelines for using HTTP with GraphQL clients and servers.
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+# { "graphiql": true }
+operation { 
+  updateHumanName(id: "1000", name: "Luke Starkiller" ) {
+    id
+    name
+  }
+}
+```
+
+The `message` key inside the `errors` object provides some helpful information for the developer to understand what kind of error was raised, and the `locations` key indicates where the error occurred in the document.
```
benjie:
```suggestion
The `message` key inside the `errors` object provides some helpful information for the developer to understand what kind of error was raised, and the `locations` key, if present, indicates where the error occurred in the document.
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+    name
+  }
+}
+```
+
+In the previous examples, we can see that when a request error occurs the `data` key will not be included because the server returns an error response to the client before the field resolvers are executed.
+
+### Field errors
+
+Field errors are raised if something unexpected happens when a resolver function is called. For example, a resolver may provide a null value for a field that has a Non-Null output type or there may be some other internal server error during execution.
```
benjie:
```suggestion
Field errors are raised if something unexpected happens during execution. For example, a resolver may raise an error directly, or may return invalid data such as a null value for a field that has a Non-Null output type.
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+In these cases, the `data` key will be included in the result with the `errors` key because GraphQL will try to continue execution beyond the problematic field and potentially return a partial result.
```
benjie: The "potentially" here makes this was a little incorrect: either the data/errors key are both present, or it's not a partial result. I've reworded it, I think this is clearer?

```suggestion
In these cases, GraphQL will attempt to continue executing the other fields and return a partial response, with the `data` key appearing alongside the `errors` key.
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+### Network errors
+
+As with network calls to any type of API, network errors that are not specific to GraphQL may happen at any point during a request. These kinds of errors will block communication between the client and server before the request is complete, such as an SSL error or a connection timeout. Depending on the GraphQL server and client libraries that you choose, there may be features built into them that support special network error handling such as retries for failed operations.
+
+## Extensions
+
+The final top-level key allowed by the GraphQL specification in a response is the `extentions` key. This key is reserved for GraphQL implementations to provide additional information about the response and there are no restrictions on what it may contain.
```
benjie:
```suggestion
The final top-level key allowed by the GraphQL specification in a response is the `extentions` key. This key is reserved for GraphQL implementations to provide additional information about the response and though it must be an object if present, there are no other restrictions on what it may contain.
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+### Network errors
+
+As with network calls to any type of API, network errors that are not specific to GraphQL may happen at any point during a request. These kinds of errors will block communication between the client and server before the request is complete, such as an SSL error or a connection timeout. Depending on the GraphQL server and client libraries that you choose, there may be features built into them that support special network error handling such as retries for failed operations.
+
+## Extensions
+
+The final top-level key allowed by the GraphQL specification in a response is the `extentions` key. This key is reserved for GraphQL implementations to provide additional information about the response and there are no restrictions on what it may contain. 
+
+For example, some GraphQL servers may include telemetry data or information about rate limit consumption under this key. Note that what data are available in `extensions` and whether this data is available in production or development environments will depend entirely on the specific GraphQL implementation.
```
benjie: The "data are" didn't read right to me... but "data" is the plural of "datum" so... it should be right? Long story short, TIL a ["singular collective mass noun"](https://en.wikipedia.org/wiki/Data_(word)) is a thing! :rofl: 

```suggestion
For example, some GraphQL servers may include telemetry data or information about rate limit consumption under this key. Note that what data is available in `extensions` and whether this data is available in production or development environments will depend entirely on the specific GraphQL implementation.
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+## Next steps
+
+To recap what we've learned about GraphQL response formats:
+
+- The GraphQL specification allows three top-level keys in a response: `data`, `errors`, and `extensions`
```
benjie, line 114:
```suggestion
- The GraphQL specification allows three top-level keys in a response: `data`, `errors`, and `extensions`
- At least one of `data` or `errors` will be present on the response (when it contains both it is a "partial response")
```

src/pages/learn/response.mdx
```diff
@@ -0,0 +1,121 @@
...
+- The `data` key contains the result of the executed operation
+- Information about raised errors is included in the `errors` key of the response
+- Request errors (such as syntax or validation errors) are raised before execution begins so there won't be any data included in the response
```
benjie:
```suggestion
- Request errors (such as syntax or validation errors) are raised before execution begins so `data` won't be included in the response
```

benjie (changes requested):
This is great content! Excellent work! :raised_hands:

benjie (approved):
I think this is good to go - excellent work :raised_hands:

src/pages/learn/execution.mdx
```diff
@@ -114,39 +120,37 @@ Human: {
...
+Notice that our type system claims `appearsIn` will return Enum types with known values, however, this function is returning numbers! Indeed if we look up at the result we'll see that the appropriate values for the Enum type are being returned. What's going on?
 
-This is an example of scalar coercion. The type system knows what to expect and will convert the values returned by a resolver function into something that upholds the API contract. In this case, there may be an Enum defined on our server which uses numbers like `4`, `5`, and `6` internally, but represents them as Enum values in the GraphQL type system.
+This is an example of _scalar coercion_. The type system knows what to expect and will convert the values returned by a resolver function into something that upholds the API contract. In this case, there may be an Enum type defined on our server that uses numbers like `4`, `5`, and `6` internally, but represents them as the expected values in the GraphQL type system.
```
eddeee888, line 125: Hi @mandiwise @benjie, this section is called Scalar coercion but an Enum is used as example. Since these are two separate types, conflating the two could lead to confusion for the readers.

I believe the spec uses the term `Result coercion` to describe the concept in this section. Maybe the following minor changes could make it clearer:

- Changing `scalar coercion` to `result coercion`
- Adding an example of the enum coercion implementation e.g.

```ts
Episode: {
  NEWHOPE: 4
  EMPIRE: 5
  JEDI: 6 
}
```

Sorry for the late comment, I just happen to see this PR after it's merged. I'm happy to make contributions if the above makes sense.

benjie: I agree the mix of scalar/enum is confusing and would love to see your suggested edits in a PR 👍

PR #1812, opened 2024-11-02
Update introspection docs

mandiwise:
## Description

This PR contains updated introspection content for the Learn docs. Key changes include:

- A new section on type name introspection
- Documentation comments in the demo schema have been turned into description strings so they may be queried in mini GraphiQL (fixes last example where `description` field returned `null`)
- Added comment on why introspection may not be available in production
- Light editing throughout the Introspection page

@benjie @jorydotcom

### DIFF
--- a/src/pages/learn/introspection.mdx
+++ b/src/pages/learn/introspection.mdx
@@ -1,23 +1,41 @@
 # Introspection
 
-It's often useful to ask a GraphQL schema for information about what
-queries it supports. GraphQL allows us to do so using the introspection
-system!
+<p className="learn-subtitle">Learn how to query information about a GraphQL schema</p>
 
-For our Star Wars example, the file
-[starWarsIntrospection-test.ts](https://github.com/graphql/graphql-js/blob/main/src/__tests__/starWarsIntrospection-test.ts)
-contains a number of queries demonstrating the introspection system, and is a
-test file that can be run to exercise the reference implementation's
-introspection system.
+It's often useful to ask a GraphQL schema for information about what operations it supports. GraphQL allows us to do so using the [introspection system](https://spec.graphql.org/draft/#sec-Introspection).
 
-We designed the type system, so we know what types are available, but if
-we didn't, we can ask GraphQL, by querying the `__schema` field, always
-available on the root type of a Query. Let's do so now, and ask what types
-are available.
+Introspection queries are special kinds of queries that allow you to learn about a GraphQL API's schema, and they also help power GraphQL development tools. On this page, we'll learn how to run different queries to learn more about an underlying schema's types, fields, and descriptions.
+
+## Type name introspection
+
+We have already seen an example of introspection on the [Schemas and Types page](/learn/schema/). When querying a field that returned Union type, we included the `__typename` meta-field directly in a selection set to get the string value of the names of the different types returned by a search query. Let's look at this example again:
 
 ```graphql
 # { "graphiql": true }
-{
+query {
+  search(text: "an") {
+    __typename
+    ... on Character {
+      name
+    }
+    ... on Starship {
+      name
+    }
+  }
+}
+```
+
+We didn't add the `__typename` field to our GraphQL API explicitly—the GraphQL specification says that it must be provided to clients by a GraphQL implementation. This field can be queried for any field with an Object, Interface, or Union type as an output type.
+
+## Schema introspection
+
+Introspection can do more than provide type names in a query. If you designed the type system for a GraphQL API, then you'll likely already know what types are available. But if you didn't design it, you can ask GraphQL by querying the `__schema` field, which is always available on the `query` root operation type.
+
+Let's do so now and ask what types are available in the Star Wars schema:
+
+```graphql
+# { "graphiql": true }
+query {
   __schema {
     types {
       name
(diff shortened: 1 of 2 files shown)


src/pages/learn/introspection.mdx
```diff
@@ -1,23 +1,41 @@
...
-queries it supports. GraphQL allows us to do so using the introspection
-system!
+<p className="learn-subtitle">Learn how to query information about a GraphQL schema</p>
 
-For our Star Wars example, the file
-[starWarsIntrospection-test.ts](https://github.com/graphql/graphql-js/blob/main/src/__tests__/starWarsIntrospection-test.ts)
-contains a number of queries demonstrating the introspection system, and is a
-test file that can be run to exercise the reference implementation's
-introspection system.
+It's often useful to ask a GraphQL schema for information about what operations it supports. GraphQL allows us to do so using the [introspection system](https://spec.graphql.org/draft/#sec-Introspection).
```
benjie: Perhaps "features" is safer, avoids confusion with operation types (query/mutation/subscription), but either way it's better than my arch-nemesis "queries" :grin: 

```suggestion
It's often useful to ask a GraphQL schema for information about what features it supports. GraphQL allows us to do so using the [introspection system](https://spec.graphql.org/draft/#sec-Introspection).
```

src/pages/learn/introspection.mdx
```diff
@@ -1,23 +1,41 @@
...
+      name
+    }
+    ... on Starship {
+      name
+    }
+  }
+}
+```
+
+We didn't add the `__typename` field to our GraphQL API explicitly—the GraphQL specification says that it must be provided to clients by a GraphQL implementation. This field can be queried for any field with an Object, Interface, or Union type as an output type.
```
benjie: This text suffers the same issue that we hit quite a lot across the ecosystem, which is that really we're referring to the "unwrapped type" being Object/Interface/Union since it's also valid on types like `[[[User!]!]!]!`. If you can come up with a natural way of solving this without being overly technical, I'd love to hear it!

src/pages/learn/introspection.mdx
```diff
@@ -146,25 +144,36 @@ tell us what these are lists of.
...
 }
 ```
 
-So we can access the documentation about the type system using introspection,
-and create documentation browsers, or rich IDE experiences.
+As demonstrated above, we can access the documentation about the type system using introspection and create documentation browsers or rich IDE experiences.
+
+This has just scratched the surface of the introspection system; we can query for Enum type values, what Interface types another type implements, and more. We can even introspect on the introspection system itself.
+
+To see an example of a specification-compliant GraphQL query introspection system implemented in code, you can view the [introspection.ts](https://github.com/graphql/graphql-js/blob/main/src/type/introspection.ts) file in GraphQL.js repository.
```
benjie: I've switched this to a permalink, though I'm not entirely sure I should... Trade-offs either way I think.

I also think we should refer to "GraphQL.js" as "the reference implementation" as it makes it clearer why we're directing people there.

```suggestion
To see an example of a specification-compliant GraphQL query introspection system implemented in code, you can view [src/type/introspection.ts](https://github.com/graphql/graphql-js/blob/e9b6b626f6f6aa379bb8f8c48df40d0c02a26082/src/type/introspection.ts) in the reference implementation.
```

src/pages/learn/introspection.mdx
```diff
@@ -146,25 +144,36 @@ tell us what these are lists of.
...
+
+## Introspection in production
+
+Introspection is a useful feature of GraphQL, especially for client developers. However, for private GraphQL APIs, it may be undesirable to expose sensitive information about the schema publicly via the `__schema` field in a production environment.
+
+In those cases, it may be preferable to disable introspection in production environments as a part of a broader API security strategy that could also include authentication and authorization, depth-limiting, query cost analysis, and more.
```
benjie: We should not promote disabling introspection as a way to prevent exposing sensitive information about the schema. The schema should not _have_ sensitive information in it. However, it is reasonable to disable introspection to reduce the attack surface. I've rewritten these two paragraphs to clarify what is meant by a "private GraphQL API" (by avoiding use of that term) and have rewritten to encourage disabling introspection in production without suggesting that doing so hides sensitive information. I've also highlighted that you can either do safe-listing or you can do a number of other techniques and hope for the best.

```suggestion
Introspection is a useful feature of GraphQL, especially for client developers and tooling. However, for APIs intended only for your own applications, it’s typically not needed in production—required operations are usually baked into these applications at build time, making runtime introspection unnecessary.

Disabling introspection in production is common in order to reduce the API’s attack surface. This is often part of a broader API security strategy, which may also include authentication and authorization, operation safe-listing (or a range of alternative protections, such as depth-limiting, breadth-limiting, alias limits, cycle rejection, cost analysis, etc.), execution timeouts, and more.
```

src/pages/learn/introspection.mdx
```diff
@@ -146,25 +144,36 @@ tell us what these are lists of.
...
+
+## Next steps
+
+To recap what we've learned about introspection:
+
+- Type names can be queried in a field selection set for an Object, Interface, or Union type using the `__typename` meta-field
+- Information about the elements of a GraphQL schema can be queried using the `__schema` field on the `query` root operation type
+- Introspection is often disabled in production environments for private APIs
```
benjie: Even public APIs sometimes disable introspection and instead provide SDL files, and some who enable introspection in production have commented on how much effort it is to support! So let's simplify this:

```suggestion
- Introspection is often disabled in production environments
```

benjie (changes requested):
I really like this editorial and expansion, especially adding a section on disabling introspection in production (though I have comments on what that section should contain :wink:)

src/pages/learn/introspection.mdx
mandiwise: We could probably just provide another example for illustrative purposes instead of getting too into the weeds with a written explanation. Maybe something like this, then call out that it can also provide the underlying type names when dealing with wrapped types:

```graphql
query {
  hero {
    name
    friends {
      __typename
      name
    }
  }
}
```

benjie: Another example would certainly clarify, but perhaps we can just insert some subtle qualifier that makes it more precise without needing additional content - clearly "unwrapped output type" is too technical but maybe we can use something like "the ultimate output type" or "the underlying output type" or similar?

```suggestion
We didn't add the `__typename` field to our GraphQL API explicitly—the GraphQL specification says that it must be provided to clients by a GraphQL implementation. This field can be queried for any field with an Object, Interface, or Union type as the underlying output type.
```

benjie (approved):
Excellent work; I think this is good to go with or without the suggestion above :raised_hands:

benjie (approved):

PR #1814, opened 2024-11-03
Update authorization and pagination docs

mandiwise:
## Description

This PR contains updated authorization and pagination content for the Learn docs. Key changes include:

- Intro and recap formats are standardized with other updated pages
- Example resolver format has been standardized with the rest of the docs
- The `postRepository` code example has been expanded for clarity
- Nod to auth directives (given that this is a common pattern)
- Schema examples for connection and edge types are include on the pagination page
- Light editing throughout each page

@benjie @jorydotcom

### DIFF
--- a/src/pages/learn/authorization.mdx
+++ b/src/pages/learn/authorization.mdx
@@ -1,52 +1,92 @@
 # Authorization
 
-> Delegate authorization logic to the business logic layer
+<p className="learn-subtitle">Delegate authorization logic to the business logic layer</p>
+
+Most APIs will need to secure access to certain types of data depending on who requested it, and GraphQL is no different. GraphQL execution should begin after [authentication](/graphql-js/authentication-and-express-middleware/) middleware confirms the user's identity and passes that information to the GraphQL layer. But after that, you still need to determine if the authenticated user is allowed to view the data provided by the specific fields that were included in the request. On this page, we'll explore how a GraphQL schema can support authorization.
+
+## Type and field authorization
 
 Authorization is a type of business logic that describes whether a given user/session/context has permission to perform an action or see a piece of data. For example:
 
 _"Only authors can see their drafts"_
 
-Enforcing this kind of behavior should happen in the [business logic layer](/learn/thinking-in-graphs/#business-logic-layer). It is tempting to place authorization logic in the GraphQL layer like so:
-
-```js
-const postType = new GraphQLObjectType({
-  name: 'Post',
-  fields: {
-    body: {
-      type: GraphQLString,
-      resolve(post, args, context, { rootValue }) {
-        // return the post body only if the user is the post's author
-        if (context.user && (context.user.id === post.authorId)) {
-          return post.body
-        }
-        return null
-      }
-    }
-  }
-})
+Enforcing this behavior should happen in the [business logic layer](/learn/thinking-in-graphs/#business-logic-layer). Let's consider the following `Post` type defined in a schema:
+ 
+```graphql
+type Post {
+  authorId: ID!
+  body: String
+}
 ```
 
-Notice that we define "author owns a post" by checking whether the post's `authorId` field equals the current user’s `id`. Can you spot the problem? We would need to duplicate this code for each entry point into the service. Then if the authorization logic is not kept perfectly in sync, users could see different data depending on which API they use. Yikes! We can avoid that by having a [single source of truth](/learn/thinking-in-graphs/#business-logic-layer) for authorization.
+In this example, we can imagine that when a request initially reaches the server, authentication middleware will first check the user's credentials and add information about their identity to the `context` object of the GraphQL request so that this data is available in every field resolver for the duration of its execution.
 
-Defining authorization logic inside the resolver is fine when learning GraphQL or prototyping. However, for a production codebase, delegate authorization logic to the business logic layer. Here’s an example:
+If a post's body should only be visible to the user who authored it, then we will need to check that the authenticated user's ID matches the post's `authorId` value. It may be tempting to place authorization logic in the resolver for the post's `body` field like so:
 
 ```js
-// Authorization logic lives inside postRepository
-const postRepository = require('postRepository');
-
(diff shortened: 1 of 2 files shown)


src/pages/learn/authorization.mdx
```diff
@@ -1,52 +1,92 @@
...
+  IS_AUTHOR
+}
+
+type Post {
+  authorId: ID!
+  body: String @auth(rule: IS_AUTHOR)
+}
+```
+
+It would be up to the GraphQL implementation to determine how an `@auth` directive affects execution when a client makes a request that includes the `body` field for `Post` type. However, the authorization logic should remain delegated to the business logic layer.
```
benjie: I'm struggling to see how this leaves the logic in the business logic layer - isn't it the GraphQL layer here that's dictating that the rule to use is `IS_AUTHOR` rather than the business logic?

src/pages/learn/pagination.mdx
```diff
@@ -129,6 +161,14 @@ To see this in action, there's an additional field in the example schema, called
 }
 ```
 
-## Connection Specification
+## Connection specification
+
+To ensure a consistent implementation of this pattern, the Relay project has a formal [specification](https://facebook.github.io/relay/graphql/connections.htm) you can follow for building GraphQL APIs that use a cursor-based connection pattern.
```
benjie:
```suggestion
To ensure a consistent implementation of this pattern, the Relay project has a formal [specification](https://relay.dev/graphql/connections.htm) you can follow for building GraphQL APIs that use a cursor-based connection pattern - whether or not use you Relay.
```

benjie (changes requested):
Some really good changes here, but I'm not sure about "Using type system directives" being included?

src/pages/learn/authorization.mdx
mandiwise: I agree, this one is a bit ambiguous. I thought it was worth mentioning the type system directive approach because I've seen it used so often (whether or not it's a great idea to mix up auth rules in a schema). But I struggled to come up with a good example here that was adjacent to the prior example and also adheres to what would be considered the best practice. If you'd rather not invite readers to consider this option, I can just remove the section. Alternatively, I could remove the example and just provide a couple sentences explaining that people may see the type system directive approach in the wild. Let me know.

src/pages/learn/authorization.mdx
```diff
@@ -1,52 +1,92 @@
...
+  return postRepository.getBody({ user: context.user, post: obj })
+}
 ```
 
-In the example above, we see that the business logic layer requires the caller to provide a user object. If you are using GraphQL.js, the User object should be populated on the `context` argument or `rootValue` in the fourth argument of the resolver.
+In the example above, we see that the business logic layer requires the caller to provide a user object, which is available in the `context` object for the GraphQL request. We recommend passing a fully-hydrated user object instead of an opaque token or API key to your business logic layer. This way, we can handle the distinct concerns of [authentication](/graphql-js/authentication-and-express-middleware/) and authorization in different stages of the request processing pipeline.
+
+## Using type system directives
+
+In the example above, we saw how authorization logic can be delegated to the business logic layer through a function that is called in a field resolver.
```
benjie: Following on from https://github.com/graphql/graphql.github.io/pull/1814/files#r1852923240

Instead of trying to fit the auth directive approach into the "business logic layer" narrative, let's explicitly call it out as an alternative - something that people do but is not the recommended way:

```suggestion
In the example above, we saw how authorization logic can be delegated to the business logic layer through a function that is called in a field resolver. In general it is recommended to perform all authorization logic in the business logic layer, however if you decide to implement authorization in the GraphQL layer instead then one approach is to use [type system dire...
```

(merge with next paragraph)

The remainder of the text in this section would also need light editorial to reflect this.

What do you think?

mandiwise: Sounds good, addressed in https://github.com/graphql/graphql.github.io/pull/1814/commits/587906af26627dce9374358c0c24b37246ebcb0b.

benjie:
We're back to 136 files changed @mandiwise; I think something went wrong with your merge/rebase

mandiwise:
@benjie ~~I tried to sync my fork's branch with source before fixing the commit history issue and something went haywire in Github and it auto-closed this PR. I can't reopen the PR, so let me know if you can do this from your end or if you'd like me to create a new branch. I should have everything we've done so far locally.~~

Update: I dug into this further was able to figure out what needed to be done for the PR to be re-openable. I was also able to remove the duplicated file changes from the source branch again.

src/pages/learn/authorization.mdx
```diff
@@ -1,52 +1,92 @@
...
+}
 ```
 
-In the example above, we see that the business logic layer requires the caller to provide a user object. If you are using GraphQL.js, the User object should be populated on the `context` argument or `rootValue` in the fourth argument of the resolver.
+In the example above, we see that the business logic layer requires the caller to provide a user object, which is available in the `context` object for the GraphQL request. We recommend passing a fully-hydrated user object instead of an opaque token or API key to your business logic layer. This way, we can handle the distinct concerns of [authentication](/graphql-js/authentication-and-express-middleware/) and authorization in different stages of the request processing pipeline.
+
+## Using type system directives
+
+In the example above, we saw how authorization logic can be delegated to the business logic layer through a function that is called in a field resolver. In general, it is recommended to perform all authorization logic in that layer, but if you decide to implement authorization in the GraphQL layer instead then this may be accommplished using [type system directives](/learn/schema/#directives).
```
benjie: Let's make it clear this isn't the only way

```suggestion
In the example above, we saw how authorization logic can be delegated to the business logic layer through a function that is called in a field resolver. In general, it is recommended to perform all authorization logic in that layer, but if you decide to implement authorization in the GraphQL layer instead then one approach is to use [type system directives](/learn/schema/#directives).
```

benjie (changes requested):

benjie (approved):
This looks good to go to me! :raised_hands:

PR #1818, opened 2024-11-05
Update caching docs

mandiwise:
## Description

This PR expands on the existing caching content for the Learn docs so that it addresses broader performance topics. Key changes include:

- New content on using `GET` to facilitate caching, the N+1 problem, performance monitoring, and more
- Page has been renamed "Performance" (with redirect from old route)

@benjie @jorydotcom


saihaj (commented):
let's keep a dedicated caching page on that page we can talk about different caching techniques used within community: Document Caching, Response Caching, Partial Query Caching.

We can talk about performance and add more things to that page. The performace page can link to the caching page. Performance is a broader topic, we can even take things like Trusted documents and refer in performance section.

src/pages/learn/performance.mdx
```diff
@@ -0,0 +1,127 @@
...
+Accept-Encoding: gzip
+```
+
+JSON is also very familiar to client and API developers, and is easy to read and debug. In fact, the GraphQL syntax is partly inspired by JSON syntax.
+
+## Performance monitoring
+
+Monitoring a GraphQL API over time can provide insight into how certain operations impact API performance and help you determine what adjustments to make to maintain its health. For example, you may find that certain fields take a long time to resolve due to under-optimized requests to a backing data source, or you may find that other fields routinely raise errors during execution.
+
+Observability tooling can provide insight into where bottlenecks exist in the execution of certain GraphQL operations by allowing you to instrument the API service to collect metrics, traces, and logs during requests. For example, [OpenTelemetry](https://opentelemetry.io/) provides a suite of vendor-agnostic tools that can be used in many different languages to support instrumentation of GraphQL APIs.
```
Urigo, line 77: Let's use this opportunity to add a section to https://graphql.org/community/tools-and-libraries/ around observability and link here to that section so people can go and see specific solutions around that

src/pages/learn/performance.mdx
```diff
@@ -0,0 +1,127 @@
...
+    # 3 more requests for each friend
+    friends {
+      name
+    }
+  }
+}
+```
+
+This is known as the N+1 problem, where the first request to an underlying data source leads to N subsequent requests to resolve the data for all of the requested fields.
+
```
Urigo, line 52: Can we add here also a mention of solutions that use the selection set and automatically translate it to the underlying data source?
Things like @benjie's Graphile that can turn selection sets into SQL queries, some of GraphQL Mesh's providers that can translate selection sets to oData queries for example
Mostly to explain the concept of query to query

Urigo (changes requested):
Great content!

I agree with @saihaj 's sentiment around leaving the caching specific page and linking to it from there (there is a lot of opportunities to expand that page and a lot of users are Googling that terms)

src/pages/learn/performance.mdx
mandiwise: @benjie Would you like this to be converted into a separate issue to be addressed at a later date?

mandiwise:
> let's keep a dedicated caching page on that page we can talk about different caching techniques used within community: Document Caching, Response Caching, Partial Query Caching.
> 
> We can talk about performance and add more things to that page. The performace page can link to the caching page. Performance is a broader topic, we can even take things like Trusted documents and refer in performance section.

@benjie I can transfer the "Globally unique IDs" content back to a dedicated Caching page and leave the rest of the content on the new Performance page. A deep dive into focused caching topics on the Caching page would need to be scoped to a separate project. Let me know how you would like to proceed.

src/pages/learn/performance.mdx
mandiwise: A comment on this has been added in the latest commit.

src/pages/learn/performance.mdx
benjie: Yes, that sounds like the right approach, thanks Mandi!

benjie:
> @benjie I can transfer the "Globally unique IDs" content back to a dedicated Caching page and leave the rest of the content on the new Performance page. A deep dive into focused caching topics on the Caching page would need to be scoped to a separate project. Let me know how you would like to proceed.

This involved a little thought... Really Globally Unique IDs aren't _just_ for performance, they're for identity which can relate to mutations and URLs and all sorts of things. From a quick scan of the existing page, I think what it was really talking about was _normalized_ caching which also relates to rendering consistency, updating data after mutations, and other topics. Definitely we should move it off of the performance page...

Saihaj is right that in addition to normalized caching (Global Object IDs or otherwise) there are other approaches to (client-side) caching and they each have their own trade-offs - but they also solve slightly different (but overlapping) problems - and I'm not sure that "caching" is even the right title for that... There's also other forms of caching that relate to GraphQL such as caching within proxies (like Cloudflare), caching of parse/validation results, and even caching in the business logic layer such as within DataLoader - and I'm not sure those would belong on the same page! They definitely seem more "performance" related to me.

For now, let's do as you propose and move Node IDs (all the way up to the end of "Alternatives") back to the "caching" page. Any additional content about alternative techniques can be proposed via future PRs. Ideally we'll come up with a better name that essentially means "GraphQL-focussed client-side data storage and synchronisation techniques" but, you know, pithier...

src/pages/learn/performance.mdx
benjie: I've [done the same](https://github.com/graphql/graphql.github.io/issues/1832) for the security PR, I leave this one for you to file.

mandiwise: Captured in https://github.com/graphql/graphql.github.io/issues/1833

mandiwise:
> For now, let's do as you propose and move Node IDs (all the way up to the end of "Alternatives") back to the "caching" page. Any additional content about alternative techniques can be proposed via future PRs. Ideally we'll come up with a better name that essentially means "GraphQL-focussed client-side data storage and synchronisation techniques" but, you know, pithier...

Addressed in https://github.com/graphql/graphql.github.io/pull/1818/commits/6f7fa7b3767a94759ba549ca46a3c3a286f328e8

src/pages/learn/performance.mdx
```diff
@@ -0,0 +1,77 @@
...
+<p className="learn-subtitle">Optimize the execution and delivery of GraphQL responses</p>
+
+At first glance, GraphQL requests may seem challenging to cache given that the API is served through a single endpoint and you may not know in advance what fields a client will include in an operation.
+
+In practice, however, GraphQL is as cacheable as any API that enables parameterized requests, such as a REST API that allows clients to specify different query parameters for a particular endpoint. There are many ways to optimize GraphQL requests on the client and server sides, as well as in the transport layer, and different GraphQL server and client libraries will often have common caching features built directly into them.
+
+On this page, we'll explore several different tactics that can be leveraged in GraphQL clients and servers to optimize how data is fetched from the API.
+
+## `GET` requests for queries
```
benjie: Let's link back to the caching page (please edit this!)

```suggestion
On this page, we'll explore several different tactics that can be leveraged in GraphQL clients and servers to optimize how data is fetched from the API.

## Client-side caching

There is a range of client-side caching strategies that GraphQL clients may implement to help not only with performance but also to ensure that you render a consistent and responsive interface to your users. [Read more about client-side caching](/learn/caching).

## `GET` requests for queries
```

src/pages/learn/performance.mdx
```diff
@@ -0,0 +1,77 @@
...
+  # 1 request for the hero
+  hero {
+    name
+    # 3 more requests for each friend
+    friends {
+      name
+    }
+  }
+}
+```
```
benjie: I'm not sure this is a good demonstration of the problem since the `hero.friends` field is still only called once; consider these resolvers:

```ts
function Query_hero(_, __, { userId }) {
  return sql`select * from heroes where id = ${userId}`;
}
function Hero_friends(hero) {
  return sql`select * from heroes where id in (
     select friend_id from friends where hero_id = ${hero.id}
  )`;
}
```

There's no N+1 problem here.

Instead, we should start with a list field, and then have a singular field within it, e.g. (not specifically this!)

```graphql
{
  # 1 fetch results in 6 records
  films {
    # This selection set (and everything in it) executes 6 times - this
    # field is called 6 times.
    mainHero {
      name
    }
  }
}
```

src/pages/learn/performance.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+## Demand control
+
+Depending on how a GraphQL schema has been designed, it may be possible for clients to request highly complex operations that place excessive load on the underlying data sources during execution. These kinds of operations may be sent inadvertently by an known client, but they may also be sent by malicious actors.
+
+Certain demand control mechanisms can help guard a GraphQL API against these operations, such as paginating list fields, limiting operation depth and breadth, and query complexity analysis. [You can read more about demand control on the Security page](/learn/security/#demand-control).
+
+## JSON (with GZIP)
+
+GraphQL services typically respond using JSON even though the GraphQL spec [does not require it](http://spec.graphql.org/draft/#sec-Serialization-Format). JSON may seem like an odd choice for an API layer promising better network performance, however, because it is mostly text it compresses exceptionally well with GZIP.
```
benjie: `deflate` is widely supported and tends to compress GraphQL better than `gzip` according to my testing: https://github.com/graphql/graphql-spec/issues/432#issuecomment-1810420044

```suggestion
GraphQL services typically respond using JSON even though the GraphQL spec [does not require it](http://spec.graphql.org/draft/#sec-Serialization-Format). JSON may seem like an odd choice for an API layer promising better network performance, however, because it is mostly text it compresses exceptionally well with algorithms such as GZIP, deflate and brotli.
```

(Ideally the below text would also be edited to reflect this.)

benjie (changes requested):
The vast majority of this is great content as-is; I've noted some recommended changes to a few small areas below.

src/pages/learn/performance.mdx
mandiwise: The idea here was a more naive implementation GraphQL wraps something like a REST API) and you're making 3 separate requests for each friend character to a `/character/{id}` endpoint to get them by their IDs. Perhaps I can add more context to the example rather than building out more types and fields in schema to illustrate this?

src/pages/learn/performance.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+At first glance, GraphQL requests may seem challenging to cache given that the API is served through a single endpoint and you may not know in advance what fields a client will include in an operation.
+
+In practice, however, GraphQL is as cacheable as any API that enables parameterized requests, such as a REST API that allows clients to specify different query parameters for a particular endpoint. There are many ways to optimize GraphQL requests on the client and server sides, as well as in the transport layer, and different GraphQL server and client libraries will often have common caching features built directly into them.
+
+On this page, we'll explore several different tactics that can be leveraged in GraphQL clients and servers to optimize how data is fetched from the API.
+
+## Client-side caching
+
+There is a range of client-side caching strategies that GraphQL clients may implement to help not only with performance but also to ensure that you render a consistent and responsive interface to your users. [Read more about client-side caching](/learn/caching).
```
mandiwise, line 13: @benjie I accepted the change but I'm not sure what the "please edit" note was in reference to here.

benjie: Just inviting an edit in case it was needed for consistency - e.g. with how you styled other links/etc throughout the docs. If you're happy with it, no change needed.

src/pages/learn/performance.mdx
benjie: In a REST API I'd expect this to be solved via a single request to something like `/users/:id/friends` - I think you might be suggesting that a request to `/users/:id` would return `{friend_ids: [...]}` and then you'd make individual requests to `/users/:id` for each friend, but I don't commonly see this in practice.

If we substitute "friends" for "repositories", then GitHub's REST API [works in the way that I'd suggest above](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-a-user), i.e. a `GET` to `/users/{username}/repos`.

So to update the previous resolvers to use REST instead, you could imagine something like:

```ts
function Query_hero(_, __, { userId }) {
  return GET`/heroes/${userId}`;
}
function Hero_friends(hero) {
  return GET`/heroes/${hero.id}/friends`;
}
```

Again, there's no N+1 problem here. I understand the desire to not expand the schema further, but I don't think your example clearly demonstrates the N+1 problem even for REST APIs.

benjie (changes requested):
I think the only remaining issue is the N+1 demonstration; then this is good to go! :raised_hands:

src/pages/learn/performance.mdx
benjie: The critical thing is that in your example the only resolver that is called more than once is the `name` resolver, and that is likely a trivial resolver. We need to select a non-trivial field _inside_ of a list field. So you can demonstrate this with the existing schema using something like:

````suggestion
```graphql
# { "graphiql": true }
query HeroWithFriends {
  # 1 request for the hero
  hero {
    name
    # 1 request for the hero's N friends
    friends {
      name
      # N planet requests - one for each of the hero's N friends
      planet {
        name
        climate
      }
    }
  }
}
```
````

Here you might have a resolver like:

```ts
function Hero_planet(hero) {
  return GET`/planets/${hero.planet_id}`
  // Or: return sql`select * from planets where id = ${hero.planet_id}`
}
```

This resolver suffers the N+1 problem because it will be called N times (once for each hero in the parent selection set).

It's not ideal that this has the extra layer, but it does avoid having to change the schema.

mandiwise: OK, I just pushed https://github.com/graphql/graphql.github.io/pull/1818/commits/19cedbd03820ca356c5308757ac783c33c3da3cf with a modified N+1. I used the `starships` field instead so we could work with the existing schema.

benjie (approved):
Looks good to me; thanks for all your hard work on this! :raised_hands:

saihaj (approved):
looks good. Sorry for the delays here

Issue #41, opened 2016-08-23
[Master] New graphql.org


leebyron:
This is a master task for all the various work necessary for the graphql.org relaunch with inline assignees by page.

- [ ] /index.html @leebyron
- /learn/
  - [ ] learn/index.html @stubailo 
  - [ ] learn/language/ @stubailo 
  - [ ] learn/type-system/ @stubailo 
  - [ ] learn/validation/ @leebyron
  - [ ] learn/execution/ @leebyron
  - [ ] learn/introspection/ @leebyron
  - Best Practices
    - [ ] learn/serving-over-http/
    - [ ] learn/authorization/
    - [ ] learn/pagination/
    - [ ] learn/versioning/
    - [ ] learn/performance/
    - [ ] learn/security/
    - [ ] learn/design-guidelines/
    - [ ] learn/using-variables/
    - [ ] learn/colocated-queries/
    - [ ] learn/client-caching/
    - [ ] learn/persisted-queries/
    - [ ] learn/code-generation/
    - [ ] learn/migrating-from-rest/
- /code/ @lacker 
- /community/
  - Links out to places for help
  - Calendar of upcoming events
  - Grid of conference talks
- /blog/ @leebyron

skedwards88:
How up to date is this, or is there somewhere else I should look for documentation needs? I'm interested in contributing to the GraphQL documentation.

carolstran:
@skedwards88 I believe that most (if not all) of the unchecked Best Practices articles are still needed, especially since they were mentioned in the [GSoD proposals](https://github.com/graphql/foundation/tree/master/mentorship/2020/gsod#1-extend-best-practices-articles-mediumhard). If you'd be up for writing one, that could be a great way to contribute!

@leebyron Would you be alright with me opening a new issue that focuses on the missing Best Practices articles? That way, anyone interested in contributing docs would have an up-to-date resource for what articles are still available.

skedwards88:
Good to know! I have some commitments on my plate now, but I will remember this for when my free time opens up again.

PR #1820, opened 2024-11-07
Add security best practices page.

mandiwise:
Addresses a todo in #41 

## Description

This PR adds a new security best practices page with content on transport layer security, demand control, and schema considerations.

@benjie @jorydotcom

### DIFF
--- a/src/pages/learn/security.mdx
+++ b/src/pages/learn/security.mdx
@@ -16,6 +16,16 @@ The same security measures that would be used for any type of API served with a 
 
 ## Demand control
 
+### Trusted documents
+
+For GraphQL APIs that only serve first-party clients, you can write operations as needed in development environments and then lock those operations so that only they may be sent by front-end apps in production. Specifically, using _trusted documents_ will allow you to create an allowlist of operations that can be executed against a schema.
+
+As a build step during development clients may submit their GraphQL documents to the server's allowlist, where each document is stored with a unique ID—usually its hash. At runtime, the client can send a document ID instead of the full GraphQL document, and the server will only execute requests for known document IDs.
+
+Trusted documents are simply _persisted documents_ that are deemed to not be malicious, usually because they were authored by your developers and approved through code review. Efforts are underway to [standardize _persisted documents_ as part of the GraphQL-over-HTTP specification](https://github.com/graphql/graphql-over-http/pull/264)—an increasing number of GraphQL clients and servers already support them (sometimes under their legacy misnomer: _persisted queries_).
+
+Keep in mind that trusted documents can't be used for public APIs because the operations sent by third-party clients won't known in advance. However, the other recommendations that follow can help defend a public GraphQL API from malicious operations.
+
 ### Paginated fields
 
 A first step toward implementing demand control for a GraphQL API is limiting the amount of data that can be queried from a single field. When a field outputs a List type and the resulting list could potentially return a lot of data, it should be paginated to limit the maximum number of items that may be returned in a single request.
@@ -132,14 +142,6 @@ By applying weights to the types and fields in a schema, you can estimate the co
 
 While the GraphQL specification doesn't provide any guidelines on implementing query complexity analysis or rate limits for an API, there is [a community-maintained draft specification](https://ibm.github.io/graphql-specs/cost-spec.html) for implementing custom type system directives that support these calculations.
 
-### Trusted documents
-
-For GraphQL APIs that only serve first-party clients, using _trusted documents_ will allow you to create an allowlist of operations that can be executed against a schema.
-
-As a build step during development clients may submit their GraphQL documents to the server's allowlist, where each document is stored with a unique ID—usually its hash. At runtime, the client can send a document ID instead of the full GraphQL document, and the server will only execute requests for known document IDs.
-
-Trusted documents are simply _persisted documents_ that are deemed to not be malicious, usually because they were authored by your developers and approved through code review. Efforts are underway to [standardize _persisted documents_ as part of the GraphQL-over-HTTP specification](https://github.com/graphql/graphql-over-http/pull/264)—an increasing number of GraphQL clients and servers already support them (sometimes under their legacy misnomer: _persisted queries_).
-
 ## Schema considerations
 
 ### Validating and sanitizing argument values
@@ -195,6 +197,6 @@ Auth-related considerations for GraphQL APIs are discussed in-depth on the [Auth
 To recap these recommendations for securing a GraphQL API:
 
 - General security considerations for a chosen transport protocol will usually apply to a GraphQL API as well
-- Depending on your requirements, demand control can be implemented in many ways in GraphQL, including pagination of list fields, depth limiting, breadth/batch limiting, rate limiting, and using trusted documents
+- Depending on your requirements, demand control can be implemented in many ways in GraphQL, including trusted documents, pagination of list fields, depth limiting, breadth/batch limiting, and rate limiting
 - A GraphQL schema can help support validation and sanitization of client-provided values
 - Disabling introspection and obscuring error information can make an API's schema less discoverable, but they will typically not be sufficient when used on their own (a document allowlist is likely to be a more effective solution)
--- a/src/pages/learn/security.mdx
+++ b/src/pages/learn/security.mdx
@@ -184,6 +184,8 @@ query {
 
 These hints can be helpful when debugging client-side errors, but they may provide more information about a schema in a production environment than we would like to reveal. Hiding detailed error information in GraphQL responses outside of development environments is important because, even with introspection disabled, an attacker could ultimately infer the shape of an entire schema by running numerous operations with incorrect field names.
 
+In addition to request errors, details about errors that are raised during field execution should be masked as they may reveal sensitive information about the server or underlying data sources.
+
 ## Authentication and authorization
 
 Auth-related considerations for GraphQL APIs are discussed in-depth on the [Authorization page](/learn/authorization/).
--- a/src/pages/learn/security.mdx
+++ b/src/pages/learn/security.mdx
@@ -1,0 +1,197 @@
+# Security
+
+<p className="learn-subtitle">Protect GraphQL APIs from malicious operations</p>
+
+As with any type of API, you will need to consider what security measures should be used to protect a GraphQL implementation's server resources and underlying data sources during request execution.
+
+Many generalizable API security best practices apply to GraphQL, especially in relation to a given implementation's selected transport protocol. However, there are GraphQL-specific security considerations that will make the API layer more resilient to potential attacks too.
+
+On this page, we'll survey potential attack vectors for GraphQL—many of which are denial of service attacks—along with how a layered security posture can help protect a GraphQL API from malicious operations.
+
+## Transport layer security
+
+The GraphQL specification does not require a specific transport protocol for requests, but [HTTP is a popular choice](/learn/serving-over-http/) for stateless query and mutation operations. For longer-lived subscription operations, WebSockets or server-sent events are often used. Whatever the chosen protocol, security considerations in the transport layer can provide an important first line of defense for a GraphQL API.
+
+The same security measures that would be used for any type of API served with a given transport protocol should typically be used for GraphQL as well. For example, when using HTTP for queries and mutations, you should use HTTPS to encrypt data, set appropriate timeout durations for requests, and, if using HTTP caching, ensure sensitive data is cached privately (or not at all).
+
+## Demand control
+
+### Paginated fields
+
+A first step toward implementing demand control for a GraphQL API is limiting the amount of data that can be queried from a single field. When a field outputs a List type and the resulting list could potentially return a lot of data, it should be paginated to limit the maximum number of items that may be returned in a single request.
+
+For example, compare the unbounded output of the `friends` field to the limited data returned by the `friendsConnection` field in this operation:
+
+```graphql
+# { "graphiql": true }
+query {
+  hero {
+    name
+    friends {
+      name
+    }
+    friendsConnection(first: 1) {
+      edges {
+        node {
+          name
+        }
+      }
+    }
+  }
+}
+```
+
+Read more about the connection-based model for paginating fields on the [Pagination page](/learn/pagination/).
+
+### Depth limiting
+
+One of GraphQL's strengths is that clients can write expressive operations that reflect the relationships between the data exposed in the API. However, some queries may become cyclical when the fields in the selection set are deeply nested: 
+
+```graphql
+query {
+  hero {
+    name
+    friends {
+      name
+      friends {
+        name
+        friends {
+          name
+          friends {
+            name
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Even when [the N+1 problem](/learn/performance/#the-n1-problem) has been remediated through batched requests to underlying data sources, overly nested fields may still place excessive load on server resources and impact API performance. 
+
+For this reason, it's a good idea to limit the maximum depth of fields that a single operation can have. Many GraphQL implementations expose configuration options that allow you to specify a maximum depth for a GraphQL document and return an error to the client if a request exceeds this limit before execution begins.
+
+For cases where a client may have a legitimate use case for a deeply nested query and it's impractical to set a blanket limit on all queries, you may instead opt for applying depth limits specifically to list fields instead, or using [rate limiting](/#rate-limiting) instead.
+
+### Breadth and batch limiting
+
+In addition to limiting operation depth, there should also be guardrails in place to limit the number of top-level fields and field aliases included in a single operation.
+
+Consider what would happen during the execution of the following query operation:
+
+```graphql
+query {
+  hero1: hero(episode: NEWHOPE) {
+    name
+  }
+  hero2: hero(episode: EMPIRE) {
+    name
+  }
+  ### ...
+  hero100: hero(episode: JEDI) {
+    name
+  }
+}
+```
+
+Even though the overall depth of this query is shallow, the underlying data source for the API will still have to handle a large number of requests to resolve data for the aliased `hero` field.
+
+Similarly, a client may send a GraphQL document with many batched operations in a request:
+
+```graphql
+query NewHopeHero {
+  hero(episode: NEWHOPE) {
+    name
+  }
+}
+query EmpireHero {
+  hero(episode: EMPIRE) {
+    name
+  }
+}
+### ...
+query JediHero {
+  hero(episode: JEDI) {
+    name
+  }
+}
+```
+
+Depending on the client implementation, query batching can be a useful strategy for limiting the number of round trips to a server to fetch all of the required data to render a user interface. However, there should be an upper limit on the total number of queries allowed in a single batch.
+
+As with depth limiting, a GraphQL implementation may have configuration options to restrict operation breadth, field alias usage, and batching.
+
+### Rate limiting
+
+Depth, breadth, and batch limiting help prevent broad categories of malicious operations such as cyclic queries and batching attacks, but they don't provide a way to declare that a particular field is computationally expensive to resolve. So for more advanced demand control requirements, you may wish to implement rate limiting. 
+
+Rate limiting may take place in different layers of an application, for example, in the network layer or the business logic layer. Because GraphQL allows clients to specify exactly what data they need in their queries, a server may not be able to know in advance if a request includes fields that will place a higher load on its resources during execution. As a result, applying useful rate limits for a GraphQL API typically requires a different approach than simply keeping track of the total number of incoming requests over a time period in the network layer.
+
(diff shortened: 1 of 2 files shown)


src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+}
+```
+
+Even when [the N+1 problem](/learn/performance/#the-n1-problem) has been remediated through batched requests to underlying data sources, overly nested fields may still place excessive load on server resources and impact API performance. 
+
+For this reason, it's a good idea to limit the maximum depth of fields that a single operation can have. Many GraphQL implementations expose configuration options that allow you to specify a maximum depth for a GraphQL document and return an error to the client if a request exceeds this limit before execution begins.
+
+For cases where a client may have a legitimate use case for a deeply nested query and it's impractical to set a blanket limit on all queries, you may instead opt for applying depth limits specifically to list fields instead, or using [rate limiting](/#rate-limiting) instead.
```
benjie: List limits should be used in general, they're much more vital than simple depth limits alone. This is how GraphQL.js protects against malicious introspection queries for example: https://github.com/graphql/graphql-js/blob/4ef0428602fe2e16fd075e1e3e51b1feca9aad38/src/validation/rules/MaxIntrospectionDepthRule.ts#L9

I've separated this out as its own recommendation, and added query complexity analysis as another option for dealing with this. (A heading will need to be added below for this link to be valid.)

```suggestion
Since nesting list fields can result in exponential increases to the amount of data returned, it's recommended to apply a separate smaller limit to how deeply lists can be nested.

For cases where a client may have a legitimate use case for a deeply nested query and it's impractical to set a low blanket limit on all queries, you may need to rely on techniques such as [rate limiting](/#rate-limiting) or [query complexity analysis](#query-complexity-analysis).
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+    name
+  }
+  hero2: hero(episode: EMPIRE) {
+    name
+  }
+  ### ...
+  hero100: hero(episode: JEDI) {
+    name
+  }
+}
```
benjie: `DataLoader` would likely absorb this and turn it into a single request to resolve the data. This is a valid concern, but the example is perhaps not the best illustration of it. Maybe something that couldn't easily be data-loaded would be better - e.g.

```suggestion
query {
  viewer {
    friends1: friends(limit: 1) { name }
    friends2: friends(limit: 2) { name }
    friends3: friends(limit: 3) { name }
    # ...
    friends100: friends(limit: 100) { name }
  }
}
```

(If accepted, this may need adjusting to fit the other example's schema, and the paragraph below will need updating too.)

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+As with depth limiting, a GraphQL implementation may have configuration options to restrict operation breadth, field alias usage, and batching.
+
+### Rate limiting
+
+Depth, breadth, and batch limiting help prevent broad categories of malicious operations such as cyclic queries and batching attacks, but they don't provide a way to declare that a particular field is computationally expensive to resolve. So for more advanced demand control requirements, you may wish to implement rate limiting. 
+
+Rate limiting may take place in different layers of an application, for example, in the network layer or the business logic layer. Because GraphQL allows clients to specify exactly what data they need in their queries, a server may not be able to know in advance if a request includes fields that will place a higher load on its resources during execution. As a result, applying useful rate limits for a GraphQL API typically requires a different approach than simply keeping track of the total number of incoming requests over a time period in the network layer.
```
benjie:
```suggestion
Rate limiting may take place in different layers of an application, for example, in the network layer or the business logic layer. Because GraphQL allows clients to specify exactly what data they need in their queries, a server may not be able to know in advance if a request includes fields that will place a higher load on its resources during execution. As a result, applying useful rate limits for a GraphQL API typically requires a different approach than simply keeping track of the total number of incoming requests over a time period in the network layer; therefore applying rate limits within the business logic layer is generally recommended.
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+_Query complexity analysis_ is one method that can be used to rate limit GraphQL requests by applying weights to the types and fields in a schema and then analyzing an incoming document to determine if the combination of fields included in its selection set exceeds a maximum allowable cost per request. If the request proceeds, the total cost of the request can be deducted from the overall query budget allocated for a specific period.
```
benjie: I think Query Complexity Analysis warrants its own section since you can use it as a simple complexity cap without needing to tie it into rate limiting. (Rate limiting has a higher infrastructure overhead, so it's worth pointing this out as an independently valuable technique to people who may not want to run that infrastructure yet.)

Since I've added the header, I've also reordered the sentence to avoid repetition, and tried to highlight that factoring this into rate limits is optional, but valuable.

```suggestion
### Query complexity analysis

By applying weights to the types and fields in a schema, you can estimate the cost of incoming requests using the technique known as _query complexity analysis_. If the combination of fields included in a request exceeds a maximum allowable cost per request, you may choose to reject the request outright. The estimated cost can also be factored into rate limiting: if the request proceeds, the total cost of the request can be deducted from the overall query budget allocated to a client for a specific time period.
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+While the GraphQL specification doesn't provide any guidelines on implementing query complexity analysis or rate limits for an API, there is [a community-maintained draft specification]((https://ibm.github.io/graphql-specs/cost-spec.html)) for implementing custom type system directives that support these calculations.
```
benjie: Typo?

```suggestion
While the GraphQL specification doesn't provide any guidelines on implementing query complexity analysis or rate limits for an API, there is [a community-maintained draft specification](https://ibm.github.io/graphql-specs/cost-spec.html) for implementing custom type system directives that support these calculations.
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+### Trusted documents
+
+For GraphQL APIs that only serve first-party clients, using _trusted documents_ will allow you to create an allowlist of operations that can be executed against a schema.
+
+As a build step during development, clients may submit their GraphQL documents to the server's allowlist to be stored using the hashed document string as an ID. At runtime, the client can send the hashed document ID instead of the full GraphQL document, and the server will only execute the request for known document IDs.
```
benjie: Trusted documents don't have to use a hash, so I've edited this for precision:

```suggestion
As a build step during development clients may submit their GraphQL documents to the server's allowlist, where each document is stored with a unique ID—usually its hash. At runtime, the client can send a document ID instead of the full GraphQL document, and the server will only execute requests for known document IDs.
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+The GraphQL specification doesn't require the use of trusted documents or indicate how they should be implemented, but there are an increasing number of GraphQL implementations that support them.
```
benjie: Since persisted documents are a transport level concern, they're out of scope of the GraphQL spec; I've updated this text to link to the persisted documents GraphQL-over-HTTP RFC which is already fairly widely implemented, explain the reason for the "trusted" in trusted documents, and outline the difference between them and persisted documents/queries. What do you think?

```suggestion
Trusted documents are simply _persisted documents_ that are deemed to not be malicious, usually because they were authored by your developers and approved through code review. Efforts are underway to [standardize _persisted documents_ as part of the GraphQL-over-HTTP specification](https://github.com/graphql/graphql-over-http/pull/264)—an increasing number of GraphQL clients and servers already support them (sometimes under their legacy misnomer: _persisted queries_).
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+Consider the following `ReviewInput` type that is used with the `createReview` mutation: 
+
+```graphql
+input ReviewInput {
+  stars: Int!
+  commentary: String
+}
+
+type Mutation {
+  createReview(episode: Episode, review: ReviewInput!): Review
```
benjie, line 160: If this isn't part of a larger example I'm not aware of, removing the `Episode` type may simplify this for the reader:

```suggestion
  createReview(episodeId: ID!, review: ReviewInput!): Review
```

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+```graphql
+# { "graphiql": true }
+query {
+  starship(id: 3000) {
+    width
+  }
+}
+```
+
+These hints can be helpful when debugging client-side errors, but they may provide more information about a schema in a production environment than we would like to reveal. Hiding detailed error information in GraphQL responses outside of development environments is important because, even with introspection disabled, an attacker could ultimately infer the shape of an entire schema by running numerous operations with incorrect field names.
```
benjie, line 187: In addition to validation errors ("request errors" in GraphQL spec terms), it would be wise to also talk about runtime errors ("field errors" in GraphQL spec terms) and error masking.

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,197 @@
...
+Depending on what kind of client application will submit and display the review data, it may be necessary to sanitize the string provided for the `commentary` field in case a user submits unsafe HTML in it.
+
+The sanitization step should take place in the business logic layer that is called during execution of the `createReview` resolver function. To surface this to frontend developers, it could also be expressed as a part of the schema using a custom [Scalar type](/learn/schema/#scalar-types).
+
+### Introspection
+
+[Introspection](/learn/introspection/) is a powerful feature of GraphQL that allows you to query a GraphQL API for information about its schema. However, GraphQL APIs that only serve first-party clients may forbid introspection queries in non-development environments.
+
+Note that disabling introspection is a form of "security through obscurity" and will not be sufficient to protect a GraphQL API by itself, but it may be used as a part of a broader security strategy to limit the discoverability of API information from potential attackers. For more comprehensive protection of sensitive schema details and user data, trusted documents and authorization should be used.
```
benjie: Disabling introspection is good to protect against introspection attacks that attempt to achieve denial of service; hopefully this edit clarifies the use case we're talking about here:

```suggestion
Note that disabling introspection can be used as a form of "security through obscurity," but will not be sufficient to entirely obscure a GraphQL API by itself. It may be used as a part of a broader security strategy to limit the discoverability of API information from potential attackers. For more comprehensive protection of sensitive schema details and user data, trusted documents and authorization should be used.
```

benjie (changes requested):
This is going to be a super valuable resource and one the community will refer to time and time again - thanks so much for your work on it! :sparkles: It reads really well and covers most of the important stuff without getting too bogged down in technical detail; I've noted a few minor issues with precision and have made a few suggestions for improvement below.

src/pages/learn/security.mdx
mandiwise: This is how it's defined in the example schema:

https://github.com/graphql/graphql.github.io/blob/1dd7f814e0223c9a2b4950cf0e2f8daaae97b92b/src/components/marked/swapi-schema.tsx#L32-L34

src/pages/learn/security.mdx
mandiwise: I think this helps provide better clarification.

src/pages/learn/security.mdx
benjie: Oh `Episode` is an enum! I thought it was an object type! This is fine then.

benjie (approved):
Looks good to me! Excited to get this out there :raised_hands: :tada:

src/pages/learn/security.mdx
mandiwise: Resolved in a new commit.

benjie (approved):

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+### Trusted documents
+
+For GraphQL APIs that only serve first-party clients, using _trusted documents_ will allow you to create an allowlist of operations that can be executed against a schema.
+
+As a build step during development clients may submit their GraphQL documents to the server's allowlist, where each document is stored with a unique ID—usually its hash. At runtime, the client can send a document ID instead of the full GraphQL document, and the server will only execute requests for known document IDs.
+
+Trusted documents are simply _persisted documents_ that are deemed to not be malicious, usually because they were authored by your developers and approved through code review. Efforts are underway to [standardize _persisted documents_ as part of the GraphQL-over-HTTP specification](https://github.com/graphql/graphql-over-http/pull/264)—an increasing number of GraphQL clients and servers already support them (sometimes under their legacy misnomer: _persisted queries_).
```
saihaj: people also call these _persisted operations_

saihaj (approved):

src/pages/learn/security.mdx
benjie: Also "stored queries" and "stored operations" (and maybe "stored documents", but I don't recall that?) have been used. I think that's too much detail to include in these docs though; "persisted queries" is sufficiently pervasive to warrant mentioning but the others can be glossed over.

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+### Query complexity analysis
+
+By applying weights to the types and fields in a schema, you can estimate the cost of incoming requests using the technique known as _query complexity analysis_. If the combination of fields included in a request exceeds a maximum allowable cost per request, you may choose to reject the request outright. The estimated cost can also be factored into rate limiting: if the request proceeds, the total cost of the request can be deducted from the overall query budget allocated to a client for a specific time period.
+
+While the GraphQL specification doesn't provide any guidelines on implementing query complexity analysis or rate limits for an API, there is [a community-maintained draft specification](https://ibm.github.io/graphql-specs/cost-spec.html) for implementing custom type system directives that support these calculations.
+
+### Trusted documents
+
+For GraphQL APIs that only serve first-party clients, using _trusted documents_ will allow you to create an allowlist of operations that can be executed against a schema.
```
Urigo: I would maybe put Trusted Documents at the start?
Mostly because of how important they are and also because its one of the biggest misconceptions around GraphQL and security?

I would add maybe something along the lines of - 
GraphQL is mostly use as a developer tool - meaning you can use queries as you develop your app and then lock your queries in production.
That means that you can lock in advance only the queries that your team needs.
Only in case you need public APIs, and can't use Trusted Documents, then you are leaving your API exposed to all these attacks and can use tools to defend against all the following attacks..

Urigo (dismissed):
Very exciting to have this content here!

I would also maybe take this opportunity to link from the article to sections of tools and libraries that are related to the concepts in the article?
We can add a section call "security" under https://graphql.org/community/tools-and-libraries/ and link it here

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,200 @@
...
+  hero(episode: JEDI) {
+    name
+  }
+}
+```
+
+Depending on the client implementation, query batching can be a useful strategy for limiting the number of round trips to a server to fetch all of the required data to render a user interface. However, there should be an upper limit on the total number of queries allowed in a single batch.
+
+As with depth limiting, a GraphQL implementation may have configuration options to restrict operation breadth, field alias usage, and batching.
+
```
iCarossio, line 132: Would be nice to add paragraph about recursive fragments :D

CleanShot 2024-11-18 at 17 34 09@2x

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,200 @@
...
+### Rate limiting
+
+Depth, breadth, and batch limiting help prevent broad categories of malicious operations such as cyclic queries and batching attacks, but they don't provide a way to declare that a particular field is computationally expensive to resolve. So for more advanced demand control requirements, you may wish to implement rate limiting. 
+
+Rate limiting may take place in different layers of an application, for example, in the network layer or the business logic layer. Because GraphQL allows clients to specify exactly what data they need in their queries, a server may not be able to know in advance if a request includes fields that will place a higher load on its resources during execution. As a result, applying useful rate limits for a GraphQL API typically requires a different approach than simply keeping track of the total number of incoming requests over a time period in the network layer; therefore applying rate limits within the business logic layer is generally recommended.
+
```
iCarossio, line 138: The perfect common example here is trying to by pass authentication server rate limiting using aliasing or batching. 

CleanShot 2024-11-18 at 17 40 07@2x

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,200 @@
...
+    friends3: friends(limit: 3) { name }
+    # ...
+    friends100: friends(limit: 100) { name }
+  }
+}
+```
+
+Even though the overall depth of this query is shallow, the underlying data source for the API will still have to handle a large number of requests to resolve data for the aliased `hero` field.
+
+Similarly, a client may send a GraphQL document with many batched operations in a request:
```
iCarossio, line 108: Is `batching` standard to GraphQL? I think it's not in the official specification: https://spec.graphql.org/October2021/ 

It's only on *some* GraphQL clients.

src/pages/learn/security.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+The sanitization step should take place in the business logic layer that is called during execution of the `createReview` resolver function. To surface this to frontend developers, it could also be expressed as a part of the schema using a custom [Scalar type](/learn/schema/#scalar-types).
+
+### Introspection
+
+[Introspection](/learn/introspection/) is a powerful feature of GraphQL that allows you to query a GraphQL API for information about its schema. However, GraphQL APIs that only serve first-party clients may forbid introspection queries in non-development environments.
+
+Note that disabling introspection can be used as a form of "security through obscurity," but will not be sufficient to entirely obscure a GraphQL API by itself. It may be used as a part of a broader security strategy to limit the discoverability of API information from potential attackers. For more comprehensive protection of sensitive schema details and user data, trusted documents and authorization should be used.
+
```
iCarossio, line 173: Since you already added considerations about `batching` which I think is not standard in GraphQL, you can also talk about `Field Suggestion` which exits in *some* GraphQL clients.

CleanShot 2024-11-18 at 17 42 31@2x

iCarossio (commented):
Some suggestions :)

src/pages/learn/security.mdx
benjie: A GraphQL API cannot have recursive fragments, as stated by the GraphQL specification:

https://spec.graphql.org/draft/#sec-Fragment-Spreads-Must-Not-Form-Cycles

src/pages/learn/security.mdx
benjie: The official specification doesn't even cover HTTP. We're in the process of specifying batching as part of the GraphQL-over-HTTP working group:

- https://github.com/graphql/graphql-over-http/pull/307
- https://github.com/graphql/graphql-over-http/pull/308

src/pages/learn/security.mdx
benjie: This is one of many many ways of extracting schema information despite introspection being disabled, I worry if we list this as an example then people will think they just need to solve that too and then they're good - they're not. Ultimately: GraphQL schemas should not be treated as secrets/if your schema is secret then it should not accept arbitrary operations (i.e. you should use trusted documents).

If you're interested, a little more information on this topic can be found in my GraphQLConf 2024 talk; here's a timestamped link to the relevant part: https://youtu.be/Ytt1_ZIlYdg?t=1267

benjie:
Thanks for taking the time to review @iCarossio :raised_hands:

src/pages/learn/security.mdx
iCarossio: Interesting thanks!

Note: seems like it’s not always implemented properly — a CVE we found with our team.

https://github.com/graphql-rust/juniper/security/advisories/GHSA-4rx6-g5vg-5f3j

src/pages/learn/security.mdx
iCarossio: "Looking over the shoulder" is a fun one ;)
But that's totally true!

And Nice talk and thanks for the various mentions 🙏

src/pages/learn/security.mdx
benjie: That definitely seems like a bug in that implementation, missing one of the core validation rules is quite a major diversion from the spec - I would expect that to be opt-in.

It would be really good to get the https://github.com/graphql-cats/graphql-cats project (or one like it) up and running again so it can be checked if an implementation is actually spec-compliant. If anyone was interested in taking this on, I'm fairly sure it would likely get approval from the [community grant program](https://graphql.org/community/foundation/community-grant/).

src/pages/learn/security.mdx
mandiwise: Addressed in the latest commit.

benjie:
I believe Uri's feedback has been addressed - the remaining action point I've filed an issue about:

- https://github.com/graphql/graphql.github.io/issues/1832

@iCarossio Please follow up with any suggested edits as PRs - ideally separate PRs for each topic so we can discuss each on its own grounds. Perhaps for some of the examples you propose it might make sense to add them via `<details>` tags so that people who are intrigued can read more but the casual reader doesn't need to be overwhelmed with detail? (This page is generally intended for someone new to the technology.)

Thanks for your feedback everyone, and for your work on this Mandi - I'm going to go ahead and merge! :tada:

benjie (approved):

PR #1830, opened 2024-11-19
Create atproto-did for Bluesky handle

tazsingh:
## Description

This is to connect the GraphQL Foundation Bluesky handle with `@graphql.org`

Essentially `https://graphql.org/.well-known/atproto-did` needs to exist and provide the value of `did:plc:a65ga6opvhd2h453vwscrvil`

I believe putting the `atproto-did` file in the `public/.well-known` directory in this repo should achieve that

Let me know if there is a more optimal way! Happy for others to take over alternatively 🙏

Thanks!

### DIFF
--- a/public/.well-known/atproto-did
+++ b/public/.well-known/atproto-did
@@ -1,0 +1,1 @@
+// Bluesky DID goes here
--- a/public/.well-known/atproto-did
+++ b/public/.well-known/atproto-did
@@ -1,1 +1,1 @@
-// Bluesky DID goes here
+did:plc:a65ga6opvhd2h453vwscrvil


saihaj (approved):

tazsingh:
Looks like Vercel's Preview has the correct file at the correct URL ✅ 

That gives me confidence that this is the correct approach to surface that file! 🙏

JoviDeCroock (approved):

benjie:
The `content-type` here is not ideal; it should be `text/plain`. Happy to ship this and see if it works, but if not we can always fall back to DNS.

```
$ curl -H 'accept:text/plain' -i https://graphql-github-io-git-fork-tazsi-478d1d-the-graph-ql-foundation.vercel.app/.well-known/atproto-did
HTTP/2 200 
accept-ranges: bytes
access-control-allow-origin: *
age: 263
cache-control: public, max-age=0, must-revalidate
content-disposition: inline; filename="atproto-did"
content-type: application/octet-stream
date: Tue, 19 Nov 2024 16:03:25 GMT
etag: "f7d8dd2ea4fe1f6fc550848d49592535"
last-modified: Tue, 19 Nov 2024 15:59:02 GMT
server: Vercel
strict-transport-security: max-age=63072000; includeSubDomains; preload
x-robots-tag: noindex
x-vercel-cache: HIT
x-vercel-id: lhr1::vxccs-1732032205418-daea3b2567a2
content-length: 33

did:plc:a65ga6opvhd2h453vwscrvil
```

tazsingh:
Thanks @benjie!

Yeah let's see 🤞 If needed, I believe we can update the `vercel.json` file to serve this with the `text/plain` `Content-Type`

PR #1869, opened 2024-12-18
conf page minor nits

saihaj:
### Summary

- Fix breakpoints for table so sponsor logos don't overflow.
- disable text selection in some places
- Unify the color of buttons
- Fix CTA button size on small screens

### DIFF
--- a/src/app/conf/2025/sponsors.tsx
+++ b/src/app/conf/2025/sponsors.tsx
@@ -50,7 +50,7 @@ function List({
   linkClassName?: string
 }) {
   return (
-    <div className={clsx("flex gap-6 flex-col md:flex-row", className)}>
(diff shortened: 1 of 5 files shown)


src/app/conf/2025/sponsors.tsx
```diff
@@ -50,7 +50,7 @@ function List({
   linkClassName?: string
 }) {
   return (
-    <div className={clsx("flex gap-6 flex-col md:flex-row", className)}>
+    <div className={clsx("flex gap-6 flex-col lg:flex-row", className)}>
```
dimaMachina:
```suggestion
    <div className={clsx("flex gap-6 max-lg:flex-col", className)}>
```

PR #1886, opened 2024-12-19
Update deps only in weekend to reduce noise

dimaMachina:


renovate.json
```diff
@@ -1,3 +1,4 @@
 {
-  "extends": ["config:base"]
+  "extends": ["config:base"],
+  "schedule": ["every weekend"],
```
dimaMachina:
```suggestion
  "schedule": ["every weekend"]
```

saihaj (approved):

PR #1907, opened 2024-12-23
style: enable prettier for mdx files

saihaj:
Not sure why we disabled prettier for `.mdx` files. I was working on some other PR and saw it wasn't working so decided to enable for everything.


dimaMachina (commented):
@benjie should we disable [proseWrap](https://prettier.io/docs/en/options.html#prose-wrap) for mdx? 

https://github.com/graphql/graphql.github.io/blob/c4306f11c3909fc0c876a8e5175a6495fc779f18/.prettierrc#L15

benjie:
Hold fire on this.

benjie:
Okay I dug up the [bug with proseWrap in MDX](https://github.com/prettier/prettier/issues/13586) / [bug report](https://github.com/prettier/prettier/issues/12649).

Yes, we must disable proseWrap in MDX if we enable prettier; generally I don't trust it with MDX files. I've taken to just un-ignoring each new MDX file I add - perhaps we should ignore certain problematic files but enable?

benjie:
@saihaj is it possible to generate the static website before and after this change and assert that there are no differences in the generated output?

saihaj:
> @saihaj is it possible to generate the static website before and after this change and assert that there are no differences in the generated output?

yes. The deploy preview we see on this PR is a version we can compare that with prod. We can build this branch and then try to `diff` between `source` branch build

benjie:
If there's no differences in the generated output then I think it's safe to merge. I'm personally fine with the default of not enabling prettier on MDX files though, historically they've had a number of issues.

PR #1935, opened 2025-01-07
Rewrite requests to `/graphql`

JoviDeCroock:
The swapi graphql demo has been broken for a while as it points to the netlify path returning a 301, this should be fixed by https://github.com/graphql/swapi-graphql/pull/232 but the link could look cleaner so instead this requests to /graphql

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -73,7 +73,7 @@ export default withLess(
       return [
         {
           source: "/swapi-graphql/:path*",
-          destination: "https://graphql.github.io/swapi-graphql/:path*",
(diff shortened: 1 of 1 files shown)


benjie:
Redirecting all paths to `/graphql` feels wrong - what if we want to serve GraphiQL assets from here? Why are the underlying links wrong?

JoviDeCroock:
Fair enough, we could keep doing what we do now. We'll need to fix the OPTIONS request though on swapi-graphql

benjie:
What URLs is this actually intending to redirect? Maybe we can add redirects for those specific paths, rather than a catchall?

PR #1955, opened 2025-02-03
Fix argument example: Update Query to request Human instead of Droid

### LINKED ISSUE #1944: Feedback for “Queries” : Arguments example
Arguments example defines `Query` to take a `droid`, but then actual query requests a `human`

2bdulra7manRea:
Closes #1944

## Description

This PR updates the argument example where the Query previously requested a droid, but the definition specified a human. The example now correctly defines the Query to request a human, aligning with its interface Human in the argument


benjie (approved):
Thanks!

PR #1976, opened 2025-03-17
Use "a" instead of "an"

greg0ire:
It is followed by a consonant sound.


saihaj (commented):
can you please sign the CLA?

saihaj (approved):

greg0ire:
Done… I had to provide my home address :confused: , was hoping I wouldn't have to for such a change

benjie:
Thanks for signing! I agree it's a bit silly, but it's very hard to draw the line between where a CLA is needed and where a change is so trivial as to be uncopyrightable, so we have a blanket policy of requiring CLA signing for all contributions. On the plus side, you can make larger changes now, having already signed the CLA!

PR #1986, opened 2025-04-17
Fix: incorrect path for the mutations page

gameofby:
## Description

Fix the incorrect path for the mutations page.


saihaj (approved):
thanks!

PR #1990, opened 2025-04-28
fix: ci checks

saihaj:
Noticed the new docs validation job was failing #1983. This PR fixes the syntax and merges all the checks in one job as multiple jobs.

CleanShot 2025-04-28 at 13 11 31@2x 
source: https://github.com/graphql/graphql.github.io/actions/runs/14702757073

cc @sarahxsanders


saihaj:
so looks like this one fails because our this CI job can't build the Conf websites since we are missing API keys on GitHub Actions CI. 

 
CleanShot 2025-04-28 at 18 05 52@2x


@sarahxsanders @benjie @Urigo options we have to unblock this
1. We ignore the `/conf/**` subpaths since for each conf version we have an API key to build. I am not even sure what the rate limits are so probably a wise idea.
2. Someone helps me get the API keys (we have these on our vercel) and configure on GitHub CI.

benjie:
I'd go with option 1 for now, seems the ~~safest~~ least likely to break CI flows :+1:

sarahxsanders:
I agree with @benjie

saihaj:
actually we can't do `1` because I can't find a way to ignore building the conf site 🥲

martinbonnin:
Looks like this is all good now. Closing this.

PR #1997, opened 2025-05-15
Add Board Meeting Summary

jorydotcom:
This is from the April meeting that was approved at today's Board meeting.

### DIFF
--- a/src/pages/blog/2025-05-15-governing-board-recap.md
+++ b/src/pages/blog/2025-05-15-governing-board-recap.md
@@ -1,0 +1,28 @@
+---
+title: 📣 April 2025 GraphQL Foundation Board Meeting Recap
+tags: ["blog"]
+date: 2025-05-15
+byline: GraphQL Governing Board
+---
+
+The GraphQL Foundation Governing Board met April 17 to discuss ongoing community efforts, technical progress, and upcoming events—read on for the highlights:
+
+### **🖥️ Website Redesign Underway**
+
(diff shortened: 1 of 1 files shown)


benjie (approved):

src/pages/blog/2025-05-15-governing-board-recap.md
```diff
@@ -0,0 +1,28 @@
+---
+title: 📣 April 2025 GraphQL Foundation Board Meeting Recap
+tags: ["blog"]
+date: 2025-05-15
+byline: GraphQL Governing Board
+---
+
+The GraphQL Foundation Governing Board met April 17 to discuss ongoing community efforts, technical progress, and upcoming events—read on for the highlights:
+
+### **🖥️ Website Redesign Underway**
```
Urigo, line 10: should maybe this section be moved to the community WG blog post?

jorydotcom: Not sure I follow - do you mean we should move this into a different blog post entirely? Or categorize this blog as community? I don't have a strong opinion, except to add that the goal of the public posts is just to share what the board talked about in its monthly call. In any case, this is the copy the board approved, and it had 1 month to do so, so we better ship it as is and ask all board members to provide feedback on the post before approval. I can also add future summaries as Draft PRs here if that would help with review!

Urigo: I think the goal of this post is to show what the foundation is doing.
We also want to highlight the work that the community WG is doing.
Putting on the blog post of the foundation board, work that the community WG is doing and not the foundation, could confuse people

PR #2016, opened 2025-06-09
new conf design — rate limit workaround

### LINKED ISSUE #2014: JSON export old confs
we should export the JSON from sched API for 2023 and 2024 confs so we do not have to go to sched API to generate these pages.

hasparus:
Closes #2014.

## Description

We now pull the Sched's data to JSON files to avoid the 30req/min rate limit.

Preview on my Vercel account: [fabulous-isomorphism-e22ige7yg-hasparus-projects.vercel.app](https://fabulous-isomorphism-e22ige7yg-hasparus-projects.vercel.app/)


.github/workflows/conference-sync.yml
```diff
@@ -0,0 +1,26 @@
...
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Sync conference data from Sched
+        run: |
+          tsx scripts/sync-sched/sync.ts --year 2025
+        env:
+          SCHED_ACCESS_TOKEN_2025: ${{ secrets.SCHED_ACCESS_TOKEN_2025 }}
```
saihaj, line 20: we don't really have this configured here so need @benjie to set this. (We have it over email thread or can share on slack)

hasparus: We can disregard it for now to unblock deploys. I ran the script locally, and this is a workflow_dispatch trigger. I'd prefer to run this on a webhook from Sched and maybe every N minutes (to avoid the rate limit), but we don't _need_ it yet.

saihaj (approved):

PR #2027, opened 2025-06-17
merge by name and social urls

hasparus:
## Description

As the speakers sometimes have multiple user accounts in Sessionize (or Sched) and we can't expect the `username` across years will be the same, we now merge the speakers during site build (_not during sync, during sync we just find equal full names but this is editable by a human, we never remove items from that list_).

image

### DIFF
--- a/src/app/conf/_api/sched-data.tsx
+++ b/src/app/conf/_api/sched-data.tsx
@@ -1,11 +1,34 @@
 import { ConferenceYear, SchedSpeaker } from "./sched-types"
+import { mergeSpeaker } from "./sched-client"
 
-const allSpeakers: SchedSpeaker[] = require("../../../../scripts/sync-sched/speakers.json")
+const speakersData = require("../../../../scripts/sync-sched/speakers.json")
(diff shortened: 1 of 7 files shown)


src/app/conf/_api/sched-data.tsx
```diff
@@ -1,11 +1,36 @@
...
+    if (speakersWithDuplicates.has(speaker.username)) {
+      return (
+        equalitySets
+          .find(set => set.includes(speaker.username))! // we will definitely find, because we checked .has above
+          .map(username => allSpeakers.find(s => s.username === username))
+          .filter(x => !!x)
+          // we prefer the data from the most recent years
+          .sort((a, b) => Math.max(...a._years) - Math.max(...b._years))
+          // and the `year` of the conference we're fetching data for
+          .concat([speaker])
```
hasparus, line 25: We merge speaker data starting from oldest to newest and then _correct_ year. (So building 2025 site next year it will effectively `Object.assign(s2023, s2024, s2025, s2026, s2025)`

saihaj (approved):

PR #2039, opened 2025-06-27
graphqlconf 2025 — schedule improvements

hasparus:
## Description

I've corrected the timestamps, displayed duration in the lightning talks (with partially colored background to make it more apparent), added [Add to calendar] to the Schedule as per the design, and added more options to [Add to calendar] in long session cards where we already had this button with just ICS support.

image
image
image
image


saihaj (changes requested):
dark mode the add calendar dropdown colours need to be updated.

CleanShot 2025-06-27 at 15 18 59

saihaj (approved):

PR #2057, opened 2025-07-24
conference page — missing OG images

hasparus:
## Description

I've added the missing OG images for the pages that aren't based on data from Sched.

image


src/app/conf/2025/code-of-conduct/opengraph-image.tsx
```diff
@@ -0,0 +1,10 @@
+import { SimpleOpengraphImage } from "../components/og-images/simple-opengraph-image"
+export {
+  generateStaticParams,
+  contentType,
+  size,
+} from "../components/og-images/simple-opengraph-image"
+
+export default SimpleOpengraphImage.bind(null, {
```
hasparus, line 8: We're using `bind` here instead of an arrow or a new function because `@types/react` doesn't seem to know we can return `Promise<JSX>` just yet.

This should be solved after migrating to React 19, but I'm fine with taking this small stylistic/readability hit just to a void dealing with React versions for now.

src/app/conf/2025/components/speaker-opengraph-image.tsx
```diff
@@ -1,374 +0,0 @@
-import type { SchedSpeaker } from "@/app/conf/2023/types"
```
hasparus, line 1: This file isn't removed, I've moved it to `og-images` and refactored the smaller components out to new files, so Git didn't catch that.

saihaj (approved):

PR #2100, opened 2025-08-28
Remove unpaginated fields from connection-style pagination schema example

### LINKED ISSUE #2099: Feedback for “Pagination” - extraneuos field on `FriendConnection`?
Hello! I like [your pagination documentation](https://graphql.org/learn/pagination/) and often recommend it to people to learn about connection-style pagination in GQL.

However, I don't understand what the `friend` field on `FriendConnection` is for - does this return an unpaginated list of all data? Perhaps it's just a typo and shouldn't be there at all?

```
type FriendsConnection {
  totalCount: Int
  edges: [FriendsEdge]
  friends: [Character]
  pageInfo: PageInfo!
}
```

jkillian:
Closes #2099

## Description

I believe both of these fields are implying un-paginated access to the friends list. Likely you wouldn't want this exposed at all if you're trying to expose an paginated API, so I removed them from the examples. Not sure if this is the original intention or not, but seemed reasonable to me?

### DIFF
--- a/src/pages/learn/pagination.mdx
+++ b/src/pages/learn/pagination.mdx
@@ -113,7 +113,6 @@ To see this in action, there's an additional field in the example schema, called
 interface Character {
   id: ID!
   name: String!
-  friends: [Character]
   friendsConnection(first: Int, after: ID): FriendsConnection!
   appearsIn: [Episode]!
 }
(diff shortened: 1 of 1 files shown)


src/pages/learn/pagination.mdx
```diff
@@ -113,15 +113,13 @@ To see this in action, there's an additional field in the example schema, called
 interface Character {
   id: ID!
   name: String!
-  friends: [Character]
```
jkillian, line 116: As per #2099 it might make sense to remove this instance of things, but keep the one below? Up to you all

jkillian:
It seems that as per #2099 there's use for both of these fields. Going to close this out.

PR #2111, opened 2025-09-05
GraphQL Locals: Update meet up details

jemgillam:
I have updated the meet-ups page. Now more official "GraphQL Locals" have been added, and useful details added to the EventCard such as date of last /next meet up. There's also an indication of if the meet up is an "Official GraphQL Local" supported by the Foundation. 

The details of the meet ups have been autogenerated from a list we already
 maintained elsewhere, this should mean it's easier to keep this page up to date in the future! 

Screenshot 2025-09-05 at 17 49 22

For now, have removed the "Past events" dropdown as it was very out of date.


benjie (approved):
Ace; massive improvement! :raised_hands:

saihaj (approved):

PR #2113, opened 2025-09-05
landing — interactive editor 

hasparus:
## Description

@martinbonnin proposed making the "How it Works" code samples interactive on Slack.

We already had CodeMirror, and I already needed to restyle it for the docs, so it required just a new schema, some refactoring and shuffling things around.

I think the idea makes sense, and lets the user experience GraphQL immediately. Even if they're non-programmers, they can maybe get a feel how the data flows and what the experience is like.

I changed the section CTA to state "Try GraphiQL", because we're already trying it out live.

image

### To-do

- [x] Write a small GraphQL Schema for this example
- [x] Double check if next/dynamic worked and we don't have CodeMirror in the main bundle
  - (I actually wrote `dynamic(import` instead of `dynamic(() => import` initially so it's good I double checked).
- [x] Add a test for this


martinbonnin (dismissed):
This is very cool 🤩 

May I suggest we make it more explicit that the page is interactive with a small comment/call to action?
Screenshot 2025-09-06 at 19 08 17

hasparus:
> This is very cool 🤩
> 
> May I suggest we make it more explicit that the page is interactive with a small comment/call to action? image

martinbonnin:
It wouldn't be hover, right? It'd be something else? I kind of liked that the comment was where the user would look and therefore impossible to miss but no strong opinion at this stage.

hasparus:
I'm changing the PR to a draft because we noticed a schema discrepancy and I need to write a small script first.

hasparus:
@saihaj don't merge it just yet, I'm gonna test it on deploy preview this morning bcs I'm not 100% sure about the cors header

jonathanawesome:
Found an issue with how the InteractiveEditor sits on top of the static bits.

https://github.com/user-attachments/assets/494fd87d-6f2b-41f2-8798-ee3bb04a79cc

hasparus:
Converted to draft to fix the mobile view bug Jon found in https://github.com/graphql/graphql.github.io/pull/2113#issuecomment-3444594270.

src/components/index-page/how-it-works/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+          />
+          <HowItWorksListItem
+            text="Get predictable results"
+            code={<HowItWorks_Result />}
+          />
+        </ol>
+        {inView && (
+          <ol
+            // this is rendered *on top* of the static version to avoid layout shift
+            className="max-lg:before pointer-events-none absolute inset-0 list-none gap-px [counter-set:list-item_1] *:pointer-events-auto max-lg:before:absolute max-lg:before:inset-x-0 max-lg:before:-top-2 max-lg:before:bottom-[98%] max-lg:before:bg-[rgb(var(--nextra-bg))] max-sm:top-[191px] sm:max-lg:top-[214px] lg:grid lg:grid-cols-3 lg:[&>:first-child]:col-start-2"
```
jonathanawesome, line 61: Should be `[counter-set:list-item_0]` and `lg:[&>:first-child]:col-start-1`.

src/components/index-page/how-it-works/interactive-editor.tsx
```diff
@@ -0,0 +1,120 @@
...
+      }}
+      runQuery={() => {
+        setVariableTypes(getVariableToType(schema, query))
+        runQuery({ manual: true })
+      }}
+    />
+  )
+
+  return (
+    <>
```
jonathanawesome, line 82: rendering the first list item here, along with the css changes above, cleans things up nicely for me. it means rendering the first list item twice, but seems like a fair trade off.

jonathanawesome (commented):
https://github.com/user-attachments/assets/3fdcf922-8e8c-4f75-8f18-7f5b797196ce

src/components/index-page/how-it-works/index.tsx
hasparus: uh but the counters are okay now, aren't they?

hasparus: even on the video where you recorded the bug

jonathanawesome: Yes, actually, I think you're right. It's possible that I hadn't pulled down your latest changes before checking, because I'm a different machine now, pulled your PR down fresh and it looks good!

hasparus:
> This is very cool 🤩
> 
> May I suggest we make it more explicit that the page is interactive with a small comment/call to action?

@martinbonnin I remember somebody else said they didn't even know the examples in the docs are interactive.

What do you think about this?

image

Alternative shorter wording:

image

martinbonnin:
@hasparus sorry I forgot who that was 😅 

```
Press CtrlSpace to open the completions menu, ⌘Enter to run the query.
```

That doesn't seem to work for me (MacBook pro ISO international)

I'd suggest adding a comment that the editor is interactive:

Screenshot 2025-11-07 at 00 21 06

Thanks for looking into this in all cases, this looks cool! 🤩

PR #2129, opened 2025-09-09
Tweak the wording in the graphql.org blog post

hasparus:
TLDR: Removed "entirely new" from the blog post.

## Description

I'm sorry, when I was reading the post before, it didn't strike me, but when I think about it now, I think we should rephrase a bit.

As you can see in its repo description (gib me a star pls :P), the theme is based on Min Light and GitHub Dark (both MIT licensed), and the parts of them that appear only in the editor, not on the website (e.g. editor controls) are _mostly_ unchanged.

I also removed the word "accidentally" because we didn't _plan_ to do it initially, it's not like an AI agent wrote `vsce publish`. I did it because after creating the theme to match designs and testing the code snippets we have in these docs, calling `vsce publish` was basically free.


src/pages/blog/2025-09-08-announcing-graphqldotorg.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Introducing the New GraphQL.org: A Decade of Evolution, Redesigned"
-tags: ["announcement"]
+tags: ["announcements"]
```
hasparus, line 3: image

👀

saihaj (approved):

PR #2166, opened 2025-10-03
Update requirements and wording of the locals page

martinbonnin:
* Move video recording to a soft requirement: photos are ok if video is not possible
* Explain why there is a requirement for a TSC representative to be an organizer

### DIFF
--- a/src/pages/community/foundation/local-initiative.mdx
+++ b/src/pages/community/foundation/local-initiative.mdx
@@ -35,7 +35,7 @@ Organizers of a Local must agree:
   **[GraphQL Brand Guidelines](https://graphql.org/brand/)**.
 - At least one
   [GraphQL Technical Steering Committee (TSC) member](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#tsc-members)
-  must be granted administrator access to the Local’s online platforms
+  or [Ambassador](https://github.com/graphql/community-wg/tree/main/rfcs/ambassador-program) must be granted administrator access to the Local’s online platforms
   - This is to ensure continuity of the community, should any of the original organizers not be available anymore.
 - The Local has 1 or more organizer from the area. (We're pretty flexible on the
   definition of "from the area", just be reasonable.)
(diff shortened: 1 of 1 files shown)


vliegveld5:
For `Explain why there is a requirement for a TSC representative to be an organizer`
Can we also leverage GraphQL Ambassadors?

Instead of only TSC representative, to include GraphQL Ambassadors as representative as part of the GraphQL local organisation?

martinbonnin:
> Instead of only TSC representative, to include GraphQL Ambassadors as representative as part of the GraphQL local organisation?

TSC is more stable IMO (there's a 2 years commitment) and shouldn't go "missing in action". I would stick with that unless we have a hard time recruiting them.

martinbonnin:
That being said, I'm not sure what the current status. Looks like most of the locals don't have a TSC representative in their org anyways so if adding Ambassadord makes it so that we actually do what we say we should do, then I'm all for it.3

martinbonnin:
Ok, I'm convinced ^^, I've added Ambassadors (ping @jemgillam )

benjie:
> That being said, I'm not sure what the current status. Looks like most of the locals don't have a TSC representative in their org

Every _official_ GraphQL Local has a TSC representative in their org, it's one of the requirements.

- London: me
- Hyderabad: me
- Ahmedabad (defunct): me
- Austin: me
- Paris: Martin
- Amsterdam: me
- Oslo: me
- Białystok: me

We're softening the requirements (e.g. for recordings) to encourage more meetups to become _official_ GraphQL Locals. We expect that many of the people who run Locals will become Ambassadors, but for the continuity of the group it's best if an official TSC member has admin access so they can add new organizers in the case of someone going non-contact. Partly it's to address bus-factor.

martinbonnin:
Right! This is the [GraphQL foundation account](), right? 

What about:

> * The local must grant the GraphQL foundation administrative access to the Local’s online platforms. This can preferably be done through one of the official GraphQL foundation account (example: [Meetup account](https://www.meetup.com/members/354034322/)) or if none is available, by inviting a [GraphQL Technical Steering Committee (TSC) member](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#tsc-members).
>    - This is to ensure continuity of the community, should any of the original organizers not be available anymore.

benjie:
The official meetup account costs something silly like £30/mo/meetup; for groups using that we cannot offer sponsorship (since the funds are already spent on meetup), so it shouldn’t be the recommended approach. Instead a TSC member should be added to whatever platform they use.

martinbonnin:
> The official meetup account costs something silly like £50/mo/meetup

Ah yes, I remember that discussion now, bummer... I've reverted to just TSC members now.

PR #2180, opened 2025-10-14
Update Events page a bit

hasparus:
## Description

Howdy,

- I updated the UI of EventCard to match designs,
- made the events scroll horizontally, as per design,
- added a scroll-timeline + mask-image based fade to the scrollview, similar to what we're using on landings, but fancier implementation as we don't have as much negative space to render the fade gradient the simpler way. This is maybe a 10th time where I used scroll-timeline, it's pretty rad, my friends, good stuff.
- added GraphQLConf 2025 to past events
- and GraphQL Day at APIDays to upcoming events.

cc @Urigo you promised me a review today on the call :)

### Differences from designs

I made the spacing in the cards a little bit smaller because we don't have as much room here in the docs as in the Events landing page they were initially designed for.

https://github.com/user-attachments/assets/5ca41ec3-3d12-4a6c-a86e-d8e2fddd4cd5

### DIFF
--- a/src/pages/community/events.mdx
+++ b/src/pages/community/events.mdx
@@ -2,65 +2,19 @@
 title: Events & Meetups
 ---
 
-{/* title can be removed in Nextra 4, since sidebar title will take from first h1 */}
-
 # Events & Meetups
 
 import { LocationIcon, ClockIcon } from "../../icons"
 import { clsx } from "clsx"
 import { useEffect } from "react"
 import { useData } from "nextra/hooks"
-import { meetups } from "../../components/meetups"
-import { events } from "../../components/events"
 import "leaflet/dist/leaflet.css"
+
+import { meetups } from "../../components/meetups"
+import { events, EventCard } from "../../components/events"
 import pinkCircle from "./pink-circle.svg"
 import { Button } from '../../app/conf/_components/button'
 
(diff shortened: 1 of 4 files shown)


martinbonnin (approved):

src/pages/community/events.mdx
```diff
@@ -93,18 +57,13 @@ export function Events({ events }) {
   )
 }
 
-{/*
-## Events
+## Upcoming Events
```
hasparus, line 60: This is a bit annoying, but we can't conditionally render this heading, because it won't landing in ToC.

Heading scanning should be handled by rehype on the final HTML, not in remark layer, but it's a detail and it won't look bad till December.

Dustin4444 (commented):
#2183

jemgillam:
What's happened here with the list of meet ups? On firefox I can only see three and there is no horizontal scrolling. Can we have tiles / cards instead? 

EDIT: Actually, I got the side scroll to work. But there is no indication it is there, it only looks like there are three meet ups on first look when trying to find your local. Looking in mobile view the tiles are also too big to fit even just one in view on eg iPhone 12 Pro. I think this experience could be improved.  

Screenshot 2025-10-21 at 16 20 55

PR #2190, opened 2025-10-16
Add GraphQL Local Initiative Update to featured posts section

hasparus:
## Description

in a response to @martinbonnin's feedback

> it's a bit weird that the first post isn't visible in the first page of the blog home

the entire concept of `featured` posts may be redundant, but for now this is you can put it above the fold by adding `featured: true` to frontmatter

I'm not sure if @benjie wanted to put it there and didn't know about the field in frontmatter or knew about the field didn't want to put it there

I do agree that displaying the newest post at the top makes sense

image

### DIFF
--- a/src/pages/blog/2025-10-20-graphql-conf-2025-article-1/index.md
+++ b/src/pages/blog/2025-10-20-graphql-conf-2025-article-1/index.md
@@ -1,0 +1,302 @@
+---
+title: "🚀 What Launched at GraphQLConf 2025"
+tags: ["blog"]
+date: 2025-10-20
+byline: Jeff Auriemma and An Ngo
+featured: true
+---
+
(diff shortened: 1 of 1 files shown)


Urigo (approved):

benjie:
I was planning to bump it to featured once the AI WG has happened and no longer needs to be front and centre. The Locals program is less time sensitive.

benjie:
But I don’t mind either way; feel free to merge 🤷‍♂️

martinbonnin:
My personal preference is that either every post is featured or we limit the number of featured posts to one so that the list of all posts becomes visible on the first page.

I get having featured articles for news site where there is a lot of daily content and you want some of it to stick around but I don't think it's our case (yet :))

martinbonnin (approved):

PR #2208, opened 2025-10-29
Enhance documentation on error handling in GraphQL

hitherejoe:
## Description

Added section on modeling errors as data in GraphQL APIs, detailing recoverable and unrecoverable errors, and how to structure mutations and queries to handle errors effectively.

This has been added based off some conversations with @Urigo and some of the team at The Guild.

A couple of notes:

- I added this as a new part of the Errors section. While the "Errors as Data" section is quite long, this felt like the best fit.
- I've adjusted the examples to match the existing operations that are used in the page, happy to tweak things more if need-be!

### DIFF
--- a/src/pages/learn/response.mdx
+++ b/src/pages/learn/response.mdx
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
 
 As with network calls to any type of API, network errors that are not specific to GraphQL may happen at any point during a request. These kinds of errors will block communication between the client and server before the request is complete, such as an SSL error or a connection timeout. Depending on the GraphQL server and client libraries that you choose, there may be features built into them that support special network error handling such as retries for failed operations.
 
+### Modelling Errors as Data
+
+When returning errors to API clients, it’s important that we enable clients to create recoverable scenarios for users. This can be enabled by providing the required context to clients, along with making it easy for clients to consume those errors.
+When modelling API errors for both queries and mutations we can follow two guiding principles:
+
+- **Unrecoverable errors, returned in the errors array**  
+  These are errors that are not the users fault (developer errors), which are generally things that the user can’t recover from. For example, the user not being authenticated or a resource not being found. This will also include scenarios such as   
+  server crashes, unhandled exceptions and exhausted resources (for example, memory or CPU)
+
+- **Recoverable errors (user errors), returned as data (typed errors)**  
+  These are errors that the user can recover from or where we need to communicate something to the user. For example, input validation error or the user having hit a limit on their plan for the requested operation. This approach allows us to     
+  utilise typed errors to provide context to clients, while allowing us to enforce a clear distinction between developer and user-facing errors. This ensures that only useful errors are returned as types and everything else developer facing will be
+  returned in the errors array.
+
+#### Mutations
+##### Modelling Errors
+Every mutation defined in the schema returns a Payload union - this allows mutations to return their success state, along with any user facing errors that may occur. This should follow the format of the mutation name, suffixed with Payload - e.g `{MutationName}Payload`.
+
+```graphql
+union CreateHeroPayload = CreateHeroSuccess | ...
+```
+
+This approach allows clients to query for the specific error types that they want to handle, as well as allow us to build a continuously evolving schema. When adding support for a new error, it should be added to the payload definition.
+
+```graphql
+union CreateHeroPayload = CreateHeroSuccess | HeroLimitReached | HeroValidationFailed
+```
+
+So that we have standardisation and flexibility in place for consuming errors, every typed error can implement the MutationError interface.
+
+```graphql
+interface MutationError {
+  message: String!
+}
+```
+
+Each error that is defined will need to implement this interface. In the example below we can see the message being implemented as part of the `MutationError` contract - alongside this, each error type may also include its own information that is specific to that error. For example, in the case of a `HeroLimitReached` error, we may want to provide the hero limit for that account so that this can be communicated to the user or for a possible error type of `HeroAlreadyExists` it would be helpful to return the `Hero` to the client.
+
+```graphql
+type HeroLimitReached implements MutationError {
+	message: String!
+	limit: Int!
+}
+```
+
+Error types can contain more complex data if they need to - for example, if we were looking to implement a `HeroValidationError` error to be returned when invalid Post data was provided during post creation, we can then model this in a way that allows multiple field errors to be handled within the composer.
+
+```graphql
+type FieldError {
+  validationError: String!
+  field: String!
+}
+
+type HeroValidationError implements MutationError {
+  message: String!
+  errors: [FieldError!]!
+}
+```
+
+When implementing the message field on the API to be returned in the response, this should be a human-readable string that can be displayed on the client. In most cases, clients will use the error type to display messaging themselves, but a default string will allow clients to display messages by default (see Future Proofing Error Responses below)
+
+##### Consuming Errors
+
+When it comes to consuming typed errors, clients can use the ... on pattern to consume specific errors being returned in the response. In some cases, clients will want to know exactly what error has occurred and then use this to communicate some information to the user, as well as possibly show a specific user-path to recover from that error. When this is the case, clients can consume the typed error directly within the mutation.
+Clients only need to consume the specific typed errors that they need to handle. For errors that fall outside of this required, the catch-all `... on MutationError` can be used to consume remaining errors in a generic fashion to communicate the given message to the user.
+
+```graphql
+mutation CreateHero {
+  createHero {
+    ... on CreateHeroSuccess {
+      // handle fields
+    }
+    ... on HeroLimitError {
+      message
+      limit
+    }
+    ... on MutationError {
+      message
+    }
+  }
+}
+```
+
+If a client does not need to consume the specific error types, they can simply rely on the `MutationError` interface:
+
+```graphql
+mutation CreateHero {
+  createHero {
+    ... on CreateHeroSuccess {
+      // handle fields
+    }
+    ... on MutationError {
+      message
+    }
+  }
+}
+```
+
+#####Future proofing error responses
+
+When mutations are first modelled in the schema it might be the case that there is not a need for any specific typed error to be defined. In future, you may add error types to the payload for a mutation, but it means that any existing clients utilising the mutation will need to update their code to consume any new errors. If you need to handle this scenario, a common mutation error type can be provided. For example, this `VoidMutationError` type will be included as a type of every mutation payload that do not include any other error types. This can then be removed in future when any user-facing error types are implemented for the payload.
+
+```graphql
+type VoidMutationError implements MutationError {
+  message: String!
+}
+
+union CreateHeroPayload = CreateHeroSuccess | VoidMutationError
+```
+
+While the API will never (and should never) explicitly return this `VoidMutationError` type, it means that when any type of MutationError is returned in future, clients will automatically receive new errors without needing to ship any changes. 
+
+```graphql
+... on MutationError {
+  message
+}
+```
+
+To benefit from this approach, client queries will need to include the resolution of the `MutationError`.
+
+##### Returning non-recoverable errors
+
+In cases where non-recoverable errors need to be returned in the errors array, our error resolver will utilise the GraphQLError class. This allows us to provide an additional code to provide more context to clients where needed. Unless we need other metadata in future, the extensions should not provide any other data outside of code that needs to be portrayed to the user - if data regarding the error is required to portray information to the user, please use a typed error.
+To enforce standards here, its good practice to define an `ErrorCode` enum which can then be provided to a function which will throw the error. This function allows you to centralise error logic and ensure that the backend is returning the correct error format for clients. Without this enforcement, it can be easy for the backend to become riddled with error codes.
+
+```graphql
+enum ErrorCode {
+  NOT_FOUND = 'NOT_FOUND',
+  FORBIDDEN = 'FORBIDDEN',
+  UNEXPECTED = 'UNEXPECTED',
+  UNAUTHORIZED = 'UNAUTHORIZED'
+}
+```
+
+```graphql
+function throwError(message: string, code: ErrorCode) {
+  throw new GraphQLError(message, { 
+    extensions: {
+      code: code,
+    },
+  });
+}
+```
+
+#### Queries
+
+In 95% of cases, GraphQL queries will only ever return either the information that was queried for, or an unrecoverable exception.
+
+```graphql
+type Query {
+  heros(input: HerosInput!): [Hero!]!
+}
+```
+
+In the above query, a successful result would see a list of Hero types returned. Otherwise, the errors array will contain any errors that were thrown during the request.
+
+However, there will be a small amount of cases where there are user-recoverable errors that may need to be returned from queries. In these cases, we should treat them the same as mutations and provide an union payload so that user-recoverable errors can be returned to the client.
+
+For example, the user could be querying for data that requires them to upgrade their plan, or to update their app. These are user-recoverable errors and utilising errors as data can improve both the Developer and User experience in these scenarios.
+
+While this is a likely to not be common when implementing queries, this approach allows us to return user recoverable errors when required.
+
 ## Extensions
 
 The final top-level key allowed by the GraphQL specification in a response is the `extensions` key. This key is reserved for GraphQL implementations to provide additional information about the response and though it must be an object if present, there are no other restrictions on what it may contain.


src/pages/learn/response.mdx
```diff
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
...
 
+### Modelling Errors as Data
+
+When returning errors to API clients, it’s important that we enable clients to create recoverable scenarios for users. This can be enabled by providing the required context to clients, along with making it easy for clients to consume those errors.
+When modelling API errors for both queries and mutations we can follow two guiding principles:
+
+- **Unrecoverable errors, returned in the errors array**  
+  These are errors that are not the users fault (developer errors), which are generally things that the user can’t recover from. For example, the user not being authenticated or a resource not being found. This will also include scenarios such as   
+  server crashes, unhandled exceptions and exhausted resources (for example, memory or CPU)
```
martinbonnin, line 112: There is also a sub-dimension: is "exceptional" vs "persistent".

`exceptional`: if the client retries, it will most likely **not** get an error.

* Out of memory error: `irrecoverable && exceptional`
* Missing resource: `irrecoverable && persistent` 

The persistent errors, you might want to model as data as well because then they become cacheable. So you don't do the same request over and over again.

src/pages/learn/response.mdx
```diff
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
...
+}
+```
+
+In the above query, a successful result would see a list of Hero types returned. Otherwise, the errors array will contain any errors that were thrown during the request.
+
+However, there will be a small amount of cases where there are user-recoverable errors that may need to be returned from queries. In these cases, we should treat them the same as mutations and provide an union payload so that user-recoverable errors can be returned to the client.
+
+For example, the user could be querying for data that requires them to upgrade their plan, or to update their app. These are user-recoverable errors and utilising errors as data can improve both the Developer and User experience in these scenarios.
+
+While this is a likely to not be common when implementing queries, this approach allows us to return user recoverable errors when required.
```
martinbonnin, line 265: What do we think of this:


```
type Query {
  hero(id: ID!): Hero # error if not found
  hero(id: ID!): HeroPayload! # typed error if not found
}
```

I think (without any data to back it up) that `hero(id: ID!): Hero` is widely more used. But this is technically recoverable so `hero(id: ID!): HeroPayload!` might make more sense?

src/pages/learn/response.mdx
```diff
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
...
+  server crashes, unhandled exceptions and exhausted resources (for example, memory or CPU)
+
+- **Recoverable errors (user errors), returned as data (typed errors)**  
+  These are errors that the user can recover from or where we need to communicate something to the user. For example, input validation error or the user having hit a limit on their plan for the requested operation. This approach allows us to     
+  utilise typed errors to provide context to clients, while allowing us to enforce a clear distinction between developer and user-facing errors. This ensures that only useful errors are returned as types and everything else developer facing will be
+  returned in the errors array.
+
+#### Mutations
+##### Modelling Errors
+Every mutation defined in the schema returns a Payload union - this allows mutations to return their success state, along with any user facing errors that may occur. This should follow the format of the mutation name, suffixed with Payload - e.g `{MutationName}Payload`.
```
martinbonnin, line 121: I'd root for `{MutationName}Result` personally but I have no idea what is used in the wild so I'd be happy to follow any existing pattern. 

Whatever we decide, we should document it and make it a lint rule. See also https://github.com/graphql/graphiql/issues/4000

martinbonnin (commented):
Super excited by this, thanks for looking into it. 

A bunch of questions. Plus I'd love to see a section about caching considerations. What is cached vs non-cached.

src/pages/learn/response.mdx
martinbonnin: Small technical nit: it's probably `{FieldName}Result` more than `{MutationName}Result` because `MutationName` usually refers to the name of the mutation in the executable document, not in the schema.

src/pages/learn/response.mdx
eddeee888: I'm curious about the "missing resource" case.

This sounds like "I query for Hero:123, and that record is missing". In this case, I'd normally return null.

Maybe I'm not thinking of the right scenario?

src/pages/learn/response.mdx
```diff
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
...
+
+```graphql
+union CreateHeroPayload = CreateHeroSuccess | HeroLimitReached | HeroValidationFailed
+```
+
+So that we have standardisation and flexibility in place for consuming errors, every typed error can implement the MutationError interface.
+
+```graphql
+interface MutationError {
+  message: String!
```
eddeee888, line 137: I'm usually on the fence about servers returning errors as string:
- If there were multiple clients (web, mobile), there's no guarantee this message would work for all of them e.g. some devices could be narrower so a long message won't work
- A client could have more specific requirement e.g. timezone, translations, locale, etc.

So, usually I find my team bypassing the `message` and using the error type (either `__typename` or `enum`). This gives more power to the client to handle errors how they want, which can be more flexible (there's pros/cons of course)

src/pages/learn/response.mdx
```diff
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
...
+  errors: [FieldError!]!
+}
+```
+
+When implementing the message field on the API to be returned in the response, this should be a human-readable string that can be displayed on the client. In most cases, clients will use the error type to display messaging themselves, but a default string will allow clients to display messages by default (see Future Proofing Error Responses below)
+
+##### Consuming Errors
+
+When it comes to consuming typed errors, clients can use the ... on pattern to consume specific errors being returned in the response. In some cases, clients will want to know exactly what error has occurred and then use this to communicate some information to the user, as well as possibly show a specific user-path to recover from that error. When this is the case, clients can consume the typed error directly within the mutation.
+Clients only need to consume the specific typed errors that they need to handle. For errors that fall outside of this required, the catch-all `... on MutationError` can be used to consume remaining errors in a generic fashion to communicate the given message to the user.
```
eddeee888, line 169: I'm curious if we should go more specific how to consume different types of errors? 
For example, I usually use `__typename`, but would mentioning something like that be too granular here?

src/pages/learn/response.mdx
eddeee888: I usually go with the `Payload` approach in queries as well. As it gives clients the ability to handle partial error.

---

For example, let's say we want to query for two sets of heroes to compare side by side on a client, I'd write the operation like this:

```graphql
query HeroCompare {
  heroSetOne: heroes(ids: ["1","2", "3", "4", "5"]) {
    id
    name
  }

  heroSetTwo: heroes(ids: "6","7", "8", "9", "10") {
    id
    name
  }
}
```

In this case:
- If there's an error in either `heroSetOne` or `heroSetTwo`, the whole query will error and we'd get `data: null`
- This could be disruptive for clients because we'd lose the ability to display the successful call

---

If we used the `Payload` approach though, if one fails (but properly handled in try/catch), the other may still complete and allows clients to handle partial failure

For example, this'd be the operation

```graphql
query {
  heroSetOne: heroesPayload(input: ["1","2","3","4","5"]) {
    ... on HeroesPayloadOk {
      result {
        id
        name
      }
    }
    ... on ResultError {
      __typename
    }
  }

  heroSetTwo: heroesPayload(input: ["6","7","8","9","10"]) {
    ... on HeroesPayloadOk {
      result {
        id
        name
      }
    }
    ... on ResultError {
      __typename
    }
  }
}
```

And if `heroSetOne` fails we'd get something like this:

```json
{
  "data": {
    "heroPayloadSetOne": {
      "__typename": "ResultError"
    },
    "heroPayloadSetTwo": {
      "result": [
        {
          "id": "6",
          "name": "Hero:6"
        },
        {
          "id": "7",
          "name": "Hero:7"
        },
        {
          "id": "8",
          "name": "Hero:8"
        },
        {
          "id": "9",
          "name": "Hero:9"
        },
        {
          "id": "10",
          "name": "Hero:10"
        }
      ]
    }
  }
}
```

eddeee888: This is also applicable to nested fields too. 

Let's say we want to build a Hero details page. A Hero has Friends and Starships. Each may trigger requests to appropriate datasources:

```graphql
type Hero {
  friends: HeroFriendsPayload!
  starships: HeroStarshipsPayload!
}
```

If either `friends` or `starships` fail, clients may still have the option to display the successful call, rather than seeing a full page error.

src/pages/learn/response.mdx
hitherejoe: @martinbonnin good point, do you mean that even in the case of a missing resource error, we would want this returned as modelled data? I can see the benefit here from caching - we previously had a NotFoundError that can be thrown in cases where a resource is not available, but I think the differentiator here would be, is it the developers fault or the users fault? We have been using typed errors for user recoverable scenarios where you need to communicate something to the user - I'm not saying that is necessarily how everyone should do it, but we were trying to find a balance between having many typed errors and keeping things simple

hitherejoe: @eddeee888 yea in the case of queries I can see that being valuable. We have been mainly modelling errors as data for mutations because we've been approaching typed errors as 'user recoverable errors'. We have had a few cases where recoverable errors have been needed for mutations, but it's not often. I think the case you have described will be valid for a lot of queries - in the case of our API users do not have direct control over the information that is being used for queries (so errors are likely our fault), so we tend to mark things as non-null and use the errors array if something goes wrong. Maybe there is a better approach to that though 😁

src/pages/learn/response.mdx
hitherejoe: Yea agreed on this, on our side clients are using the typed error to build the error message themselves. We need to enforce some type in this interface though to avoid an empty interface. I think in the actual typed errors contextual information will be provided, but we would still need to declare something in the `MutationError` 🤔

src/pages/learn/response.mdx
hitherejoe: I think there's no harm in extra clarity here,  I can add some extra examples here - `__typename` is a common approach so would likely be helpful!

src/pages/learn/response.mdx
hitherejoe: @martinbonnin for an example of our case, we have a "channelCreate" mutation, so we have a corresponding "ChannelCreatePayload". The reason for using payload instead of result was because our payload only ever contains the success state, and I felt like "Result" implies either a success or failure state (maybe I am being pedantic there though 😅)

Just checking, `{FieldName}Result` would still imply "ChannelCreateResult"? Just want to make sure I am following 😁

src/pages/learn/response.mdx
hitherejoe: I would also lean on the Payload approach here - I personally prefer this in most cases anyway as it keeps things extensible for the future. It adds a small amount of overhead for the client (having to resolve the payload type and then the result, as opposed to directly getting back the result), but feels like a better long-term approach to allow for change

I agree that `hero(id: ID!): Hero` is more widely used and something came up as a want from our FE team to keep things simpler. While it could be a preference, I think we could communicate here the benefits of using the payload approach and give reasons for leaning it it

martinbonnin: What about modeling the absence of a resource as a (semantic null):

```graphql
type Query {
  """
  returns the track for this id or null if none is found.
  """f
  track(id: ID!): Track
}
```

This is the simplest.

`Payload` on the other hand gives more room for evolution. The day a track gets a geographical restriction or a copyright strike, it's easy to add more types:

```graphql
union TrackPayload = TrackSuccess | TrackNotFound | TrackIsNotAvailableInYourCountry | TrackIsCopyrighted
type Query {
  track(id: ID!): TrackPayload
}
```

I can see room for both cases TBH. It's a conciseness vs evolvability tradeoff and different use cases may want both.

src/pages/learn/response.mdx
martinbonnin: > In this case, I'd normally return null.

Right 👍 See also [my comment below](https://github.com/graphql/graphql.github.io/pull/2208/files#r2503454259). I think we need both `null` and `union`.

> we tend to mark things as non-null and use the errors array if something goes wrong

Yes! Also, you might be interested in https://github.com/graphql/graphql-spec/pull/1163 to disable error propagation in those cases.

> I think the differentiator here would be, is it the developers fault or the users fault?

Excellent question. I would say it's the user fault. They gave a wrong id. They can recover by navigating to another Track/Product/etc... If they keep going to the same missing product page, I don't want to do a network request every time. So probably model as `null` in this case or an `union` if evolution is a concern (see also still that same other discussion)

src/pages/learn/response.mdx
martinbonnin: > The reason for using payload instead of result was because our payload only ever contains the success state,

I'm confused by the lines just below then. That seems the payload can also be an error? 


```graphql
union CreateHeroPayload = CreateHeroSuccess | HeroLimitReached | HeroValidationFailed
```



> Just checking, `{FieldName}Result` would still imply "ChannelCreateResult"

Right. The mutation root field name is `createChannel`:

```graphql
union ChannelResult = ChannelSuccess | BadId | NotAllowed

type Mutation {
  createChannel(id: ID!): ChannelResult
}
```

A "mutation name" would be there:

```graphql
mutation CreateChannel {...}
```

It's technically the same value here but in the language, "mutation name" and "field definition name" are very different AST nodes

src/pages/learn/response.mdx
eddeee888: > I can see room for both cases TBH. It's a conciseness vs evolvability tradeoff and different use cases may want both.

Agreed @martinbonnin ! Users should be empowered to pick the pattern that works best for their use case. Even teams in the same Graph may have different preferences, and that's ok IMO.

I feel if we are able to call it out it in this section, it'd really help readers as they are not forced to go with any pattern. Rather, they are presented with battle-tested patterns that could help them on their GraphQL journey.

src/pages/learn/response.mdx
eddeee888: Ah! I've just realised my variant is slightly different. 

We have a generic error type that does not implement an interface, and we use it to model error in the request (recoverable/retryable or not, they are all modelled like this).
Since we use an enum to declare the type of error, we don't have to add a message field.

For example:

```graphql
type ResultError { # generic error type
  error: ResultErrorType!
}

enum ResultErrorType {
  NOT_FOUND # This is not the equivalent of semantic null, this covers things like missing backend endpoints
  INPUT_VALIDATION_ERROR # Similar to 400
  FORBIDDEN_ERROR # Similar to 403
  UNEXPECTED_ERROR # Similar to 500
}

type Query {
  book(id: ID!): BookResult!
}

type BookResultOk {
  result: Book
}

union BookResult = BookResultOk | ResultError
```

eddeee888:
Hi @martinbonnin @hitherejoe, 

This topic is quite deep and there are a wide range of options, which requires some level of familiarity with GraphQL.

I feel the concepts may get lost in the "Learn" section (assuming this is the "The basics").
Would the "Best Practices" be a better place for an advanced concept like this?

martinbonnin:
> Would the "Best Practices" be a better place ?

Agree 100%

hitherejoe:
> Would the "Best Practices" be a better place ?

@eddeee888 @martinbonnin agreed on moving this, i'll get that adjusted. It does feel like this section is quite deep, especially with the nested headings 😀

src/pages/learn/response.mdx
```diff
@@ -102,6 +102,168 @@ The mutation above attempts to delete two starships in a single operation, but n
...
+      // handle fields
+    }
+    ... on MutationError {
+      message
+    }
+  }
+}
+```
+
+#####Future proofing error responses
```
72636c, line 203:
```suggestion
##### Future proofing error responses
```

benjie:
Hi folks! First, thanks so much @hitherejoe for this submission, it's really great to start making some recommendations in this area via the GraphQL website. Secondly, thanks all for the discussion around this! As can be seen, it is quite a subtle topic with lots of people having subtly (or sometimes radically) different opinions on it...

I started by writing my own opinions into this GitHub comment but they got quite long (it's a complex topic), so instead I've broken my thoughts out into an article on my blog:

https://benjie.dev/graphql/errors

In general I find myself agreeing with @hitherejoe; but with some nuance (see my article for definitions of the terms I'm using)... Here are a summary of my opinions:

1. The **only**[^1] valid place to model errors in the schema is as part of mutation fields (whether via a union return type or details on the mutation result payload)
2. I categorize errors differently, instead of recoverable and non-recoverable, I refer to "domain errors" vs "exceptions".
3. **Domain errors** should be modelled on mutation results (only), though using GraphQL errors for this is also acceptable if you're happy with the trade-offs.
4. **Exceptions** should never be modelled in the schema.
5. `null` should **always** be used for "does not exist" when looking up an individual resource
6. Permission denied _in a query_ is equivalent to non-existence - the resource you're attempting to request does not exist _within the set of resources you're allowed to access_. Thus the only time where NOT_FOUND or FORBIDDEN errors should be raised would be in mutation fields, and even then it should respect the principle that it should not reveal whether or not a resource you're not allowed to access exists (i.e. if you're not allowed to know the resource exists, `NOT_FOUND` should be thrown, rather than `FORBIDDEN`).
7. A union return type for mutation fields is not a one-size-fits-all solution; a mutation payload with union field on the result is more flexible at the cost of being less explicit. We should not dictate the "best practice" here when it has known issues, instead teams should make their own decisions.

For the nuances of my thoughts, please read my article :) I'll probably update it over the next few weeks as I frequently do with these articles since I wrote it in a rush.

[^1]: Okay not strictly the only place, but I don't agree 95%... I'd say it's more than 99.5% which is basically enough for us to say "only" and for people who know better to be able to justify their decisions.

martinbonnin:
For the record, @michaelstaib pointed me to this interesting post from @xuorig : https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors

It's from 2020 and provides a good background on the `FooPayload` vs `FooResult` conventions as well as many other things. 

@xuorig, if you're still around, how much of it is still valid in 2025? Anything you'd change with 5 years of feedback?

xuorig:
Hi @martinbonnin 👋 

I believe all of it is still very relevant / valid. My 2 cents, I would focus on overarching best practices rather than specific patterns, since domains/clients/server implementations can often make certain patterns easier or harder.

Things like:

-  Having wrapper types for evolution (probably the one I'd be most confident about recommending)
-  Dev-facing errors in errors array vs domain errors in the schema. (Although in real life there are real challenges to this, so I've softened my stance on this over the years)

Side-note: I would not use the wording unrecoverable vs recoverable to describe what goes into the errors array vs the schema. Both exceptions and "expected" errors can potentially be recoverable (by retrying for example)

martinbonnin:
Heads up an `error-handling` page was added as part of https://github.com/graphql/graphql.github.io/pull/2279 in the best-practices.

It's considerably less ambitious than this PR but also a good starting point if anyone feels like building on top of it.

PR #2206, opened 2025-10-25
Add head properties needed for twitter card

IestynGage:
## Description

Occasioanlly it didn't seem to like the "normal" open graph properties when playing around with a personal website, perhaps a follow up PR is needed to re-declare the og properties to work (e.g. `twitter:description`)? Although it could also be a issue with twitter/x caching things.

Added a twitter card type + set the image using a full path. The are two types of cards

summary

Image

or

summary_large_image

Screenshot 2025-10-25 at 18 51 48

I went with the first one. The image doesn't look correctly formatted however after https://github.com/graphql/graphql.github.io/issues/1963 it might just resolve itself...

Closes #2196

### Supplementary material

* https://cards-dev.x.com/validator
* https://developer.x.com/en/docs/x-for-websites/cards/overview/markup

PR #2209, opened 2025-10-30
update twitter card logo

hasparus:
Updated for logo form twitter summary cards, before we have fully generated OG images (need to do move some things into app router first / alternatively we could use an OG API Route, but it's always slow work).

We show the small card with logo only if there is no `image` for the page.


benjie:
Merged into https://github.com/graphql/graphql.github.io/pull/2206

PR #2220, opened 2025-11-06
Update the application process for the Ambassadors program

martinbonnin:
I was under the impression that there were deadline to apply and didn't realize that the applications that do not get a decision in time are rolled over. I think this was mainly because I did not get the table header and I thought the "Application Openining" period was the first column:

Screenshot 2025-11-06 at 13 06 53

This is an attempt to clarify this. @benjie @jemgillam please feel free to edit as you see fit.

### DIFF
--- a/src/pages/community/ambassadors/apply.mdx
+++ b/src/pages/community/ambassadors/apply.mdx
@@ -123,10 +123,13 @@ clarity around what and how we create a safe and collaborative environment. See
 
 ## Application Process
 
-Applications are open year-round and are received by a small panel of Foundation
-and TSC members. Once the Ambassadors program is fully up-and-running,
-Ambassadors will also have the option to review applications. The Ambassador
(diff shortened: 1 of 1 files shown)


jemgillam:
I did struggle with the table when I first wrote this. I didn't really want to put a table heading at all but I think I had to. 

I don't think this suggestion is quite right either. 

Maybe it should look more like this: https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#election-dates

jemgillam:
How about 


Screenshot 2025-11-06 at 12 53 15

martinbonnin:
> How about

I like it! One small suggestion, change `Review date` by `Review period` since it spreads several days?

martinbonnin:
Closing in favour of https://github.com/graphql/graphql.github.io/pull/2221

PR #2227, opened 2025-11-10
fix missing mobile menu in 404 page

hasparus:
## Description

- Removed MenuProvider wrapper and implement custom useMenu hook with useSyncExternalStore
- Added missing sidebar to main layout for 404 page
- Patched Nextra Theme Docs again to show the same sidebar and use the same layout wrapper in all pages in Pages router (previously Nextra used the default wrapper for layout=raw, so we had two subtly different sidebars)
- Fixed "Registeration" typo

### DIFF
--- a/src/nextra-theme-docs.css
+++ b/src/nextra-theme-docs.css
@@ -2958,3 +2958,12 @@ kbd._border._gap-1 {
     color: hsl(var(--color-neu-500));
   }
 }
+
+:root {
+  --nextra-bg: 251, 251, 249;
+}
+@media (prefers-color-scheme: dark) {
+  :root {
+    --nextra-bg: 15, 15, 12;
+  }
+}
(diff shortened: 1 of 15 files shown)


src/nextra-theme-docs.css
```diff
@@ -2958,3 +2958,12 @@ kbd._border._gap-1 {
...
 }
+
+:root {
+  --nextra-bg: 251, 251, 249;
+}
+@media (prefers-color-scheme: dark) {
+  :root {
+    --nextra-bg: 15, 15, 12;
+  }
+}
```
jonathanawesome: App-router and pages-router routes seem to be happy with this simple change.

```suggestion
.dark {
    --nextra-bg: 15, 15, 12;
}
```

jonathanawesome (changes requested):

jonathanawesome (approved):

PR #2230, opened 2025-11-12
New events and meetups

hasparus:
## Description

Refreshed the events page according to Figma.

I'll do a follow up PR with the `Add an event`, issue template, and possibly "load more" instead of a scrollview, because this is big enough already and I'd rather keep the convo about the event proposal issue template separate from the code review of this page.

https://github.com/user-attachments/assets/c6c38a21-f30b-4afc-85b3-1f65f1a04991


hasparus:
The PR is ready to review, I'm just fighting with Playwright to make sure the tests that run for me locally can also run on CI, so don't merge this yet.

src/app/(main)/community/events/meetups-map.tsx
```diff
@@ -0,0 +1,201 @@
...
+        } as React.CSSProperties
+      }
+    >
+      <div
+        className="group/map relative grow border-neu-200 bg-[--sea] dark:border-neu-50 dark:bg-[--sea] md:border-l"
+        onPointerMove={handlePointerMove}
+        onPointerLeave={handlePointerLeave}
+        onClick={handleMapClick}
+      >
+        <canvas
```
hasparus, line 163: Jon asked me a good question in DMs, so I'll note it here for posterity.

> This WebGL stuff is complicated. Why didn't you do it in the DOM?

I did initially, but it turns out that rendering _a lot_ of squares in the DOM is sorta expensive and didn't feel great on my laptop. I started getting into virtualization and reusing divs, but realized this can be written with a shader. A simpler implementation with a static image and an overlay with markers exists, but we kinda need zooming because a lot of the meetups are very close to each other.

jonathanawesome (approved):
Whoa, that's a lot of new code!

I haven't been through all of the map code in detail, but it's performant in the preview and all of the meetup nodes are generally in the correct location.

PR #2254, opened 2025-12-05
Extract both LookingForMore instances to one component

hasparus:
## Description

The designs have this section in a couple of places (some not implemented yet), so I extracted it to ensure consistent styles.


hasparus:
Subsumed by my next big PR so we can close.

PR #2262, opened 2025-12-11
Bump Next.js to latest 14.x patch

hasparus:
## Description

Updated Next.js


package.json
```diff
@@ -59,7 +59,7 @@
     "leaflet": "^1.9.4",
     "lucide-react": "^0.469.0",
     "motion": "^12.11.0",
-    "next": "^14.2.32",
+    "next": "14.2.34",
```
benjie, line 62: Do we not want a carat range for this? `^14.2.34`

PR #2278, opened 2025-12-22
Fixed typo in name

dariuszkuc:
Fixed typo

### DIFF
--- a/src/components/info-card/ambassador-data.tsx
+++ b/src/components/info-card/ambassador-data.tsx
@@ -527,7 +527,7 @@ export const ambassadors202512: Ambassador[] = [
   {
     label: "Derek Kuz",
     imageUrl: "https://github.com/dariuszkuc.png",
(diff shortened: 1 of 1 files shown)


saihaj (approved):

src/components/info-card/ambassador-data.tsx
```diff
@@ -527,7 +527,7 @@ export const ambassadors202512: Ambassador[] = [
   {
     label: "Derek Kuz",
```
benjie, line 528: @dariuszkuc Should this also have been updated?

jemgillam: Yes actually I should think it should be updated,

Sorry for that typo @dariuszkuc!

dariuszkuc: indeed -> https://github.com/graphql/graphql.github.io/pull/2289

PR #2288, opened 2026-01-05
Add Philly GraphQL

hasparus:
## Description

image
image

### DIFF
--- a/src/components/meetups.ts
+++ b/src/components/meetups.ts
@@ -285,6 +285,21 @@ export const meetups = [
       official: false,
     },
   },
+  {
+    node: {
+      id: "philly",
+      name: "Philly GraphQL",
+      latitude: 39.9526,
+      longitude: -75.1652,
+      city: "Philadelphia",
+      state: "PA",
+      country: "USA",
+      link: "https://luma.com/phillygraphql",
+      prev: "",
+      next: "2026-01-22",
+      official: false,
+    },
+  },
 ].sort((a, z) =>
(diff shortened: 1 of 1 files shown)


src/components/meetups.ts
```diff
@@ -285,6 +285,21 @@ export const meetups = [
...
+      name: "Philly GraphQL",
+      latitude: 39.9526,
+      longitude: -75.1652,
+      city: "Philadelphia",
+      state: "PA",
+      country: "USA",
+      link: "https://luma.com/phillygraphql",
+      prev: "",
+      next: "2026-01-22",
+      official: false,
```
jemgillam:
```suggestion
      official: true,
```

martinbonnin (approved):
🎉

martinbonnin:
Just a heads up I was looking for it and was a bit surprised to not see it on the event page. I had to scroll past a bit of working groups.

Screenshot 2026-01-08 at 10 27 42

Not a huge deal but for the general audience visiting graphql.org, would be nice to have some way to surface those community events a bit more.

jemgillam:
I have a feeling we said that the working groups is toggled off by default on first visit, though I guess the way this is set is then remembered by your browser!

martinbonnin:
Just tried in a private window and I get both working groups and meetups.

benjie:
Yeah I was worried about working groups basically hiding all other events. Seeing them should be explicitly opt-in on each view - I wouldn't even remember it.

PR #2291, opened 2026-01-07
Reduce internal layout jitter and make sidebar closing look better

hasparus:
## Description

This one is for the design engineers.

I changed the Collapse so the child items keep their width when the sidebar is closed and the text doesn't jump.

### Before

https://github.com/user-attachments/assets/c336956a-8a5c-46f0-a455-0244a7d35126

### After

https://github.com/user-attachments/assets/d27f7b55-9024-4152-8cb4-fc0d2de19272


jonathanawesome (approved):
wonderful addition!

PR #2292, opened 2026-01-08
Update KGraphQL

stuebingerb:
## Description

Updates KGraphQL to the current Github repository, as aPureBase has discontinued maintenance (cf. https://github.com/aPureBase/KGraphQL?tab=readme-ov-file#disclaimer).


martinbonnin (approved):
Thanks!

For the record, this is acknowledged by the aPureBase team, I think we're all good to merge ✅

Image

saihaj (approved):

PR #2307, opened 2026-01-17
Use the Button component in SME post

hasparus:
## Description

Hey @jemgillam I noticed the contrast on the links in the SME posts in light mode is a bit low (i.e. below WCAG 1.4.3 Criterion).

image

### Solution

I'd like to propose using the same button `Button` we use on the landing pages in the blogs, like this:

```tsx
import { Button } from "@/app/conf/_design-system/button";

<Button href="https://*" /> // renders HTMLAnchorElement with `target="_blank" rel="noreferrer"` 
```


jemgillam:
Good catch! Cheers!

PR #2310, opened 2026-01-19
Remove "all rights reserved"

martinbonnin:
The foundation holds the copyright and the CLA has a copyright grant clause so I think it's correct to have the foundation as a copyright owner (disclaimer: not a lawyer!).

But I don't think it's correct to say "all rights reserved". This repo is licensed under MIT and anyone has rights to use it/distribute it according to the terms of the license (again: not a lawyer!).

Also looks like this sentence is slowly becoming obsolete (see https://en.wikipedia.org/wiki/All_rights_reserved)

### DIFF
--- a/src/app/blog/rss.xml/route.ts
+++ b/src/app/blog/rss.xml/route.ts
@@ -53,7 +53,7 @@ export async function GET() {
     site_url: SITE_URL,
     // managingEditor: "info@graphql.org",
     // webMaster: "info@graphql.org",
-    copyright: `Copyright © ${new Date().getFullYear()} The GraphQL Foundation. All rights reserved.`,
+    copyright: `Copyright © ${new Date().getFullYear()} The GraphQL Foundation.`,
     language: "en-US",
     // Published date as last blog post date
     pubDate: blogs[0].frontMatter.date.toUTCString(),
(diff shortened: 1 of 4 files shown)


benjie (commented):
This will need a review from the legal team; cc @jorydotcom @leebyron

jorydotcom:
@martinbonnin the copyright notice on the graphql.org web pages should contain the "All Rights Reserved" language - this is because we have a lot of different copyrightable material on our website for which the Foundation has a license to use, but for which we do not hold the copyrights. "All Rights" are reserved by the copyright holder, so additional use (for example, reusing photos someone contributes to a blog post for a company website) would require additional permission.

martinbonnin:
@jorydotcom that makes sense to me. 
Should we update the GitHub repo then ([currently MIT](https://github.com/graphql/graphql.github.io/blob/source/LICENSE)) ? For an example, Tailwind is explicit that [their website is not open source](https://github.com/tailwindlabs/tailwindcss.com?tab=readme-ov-file#license)

benjie:
Got the answer on this: MIT license only covers software and documentation, it does not cover images, videos, or other embedded resources - hence all rights reserved.

martinbonnin:
Thanks @benjie

PR #2322, opened 2026-01-29
blog: Extend SME deadline

jemgillam:
Another small tweak to the SME timeline. 

I have also removed `featured: true` from an older AI working group post and added it to the GraphQL Conf post to bump it for visibility, I hope that's ok


martinbonnin:
Was going to suggest making the GraphQLConf post featured 👍👍

PR #2328, opened 2026-02-05
Overhaul categorization on events page

benjie:
Currently the Google Calendar is being treated by the website as returning only "(technical) working groups"; however it also contains foundation working sessions, board meetings, and other relevant events (including Locals and events that GraphQL has a presence at, such as APIDays).

This PR:

- overhauls the categorization logic by heuristically recognizing the calendar titles and bucketing them as such
- splits out "foundation meetings" from "working groups" (but does not add any styles for the former)
- shows "major" events before "minor" events - i.e. conferences and locals before working groups/foundation meetings
- changes the link for calendar events to be the event location (if it's a URL) or the first link in the description otherwise
- changes the location for calendar events to be from the calendar
- expands the visibility window for working groups to 35.5 days so we can be sure at least one of each type is shown
- applies the visibility window to foundation meetings and working groups, but not other events on the calendar (e.g. conferences, locals, etc)
- excludes GraphQLConf coming from the calendar to avoid duplicating manual definition
- shows the tags in a consistent order (conferences > meetups > working groups > foundation meetings)


src/app/(main)/community/events/events.ts
```diff
@@ -12,6 +12,14 @@ export interface Event {
...
 
 export const events: Event[] = [
+  {
+    name: "GraphQLConf 2026",
+    slug: "graphql-conf-2026",
+    location: "Menlo Park, California",
+    date: "2026-05-06T16:00:00+00:00",
+    eventLink: "/conf/2026",
+    host: "GraphQL Foundation",
+  },
```
martinbonnin, line 22: Should we put those in the calendar as well? Source of truth for everything?

martinbonnin (approved):
Looks great! Thanks! Just one minor comment.

src/app/(main)/community/events/events.ts
benjie: It's already in the calendar; but by defining it in source code we can be more specific about exactly what we want to show for these (rare) events. Going forward, all non-GraphQLConf events should come from the calendar instead IMO.

The only thing the code rejects as a duplicate is GraphQLConf instances coming from the calendar - not other conferences:

https://github.com/benjaie/graphql.github.io/blob/171d10cccd42282129067107dcd4453173d8271d/src/app/(main)/community/events/events-list.tsx#L94-L96

martinbonnin: Makes sense! Thanks for the walkthrough!

benjie:
Discussed at today's GraphQL Foundation Working Session (previously: Marketing & Content Committee) and approach was approved.

cc @hasparus for visibility since this changes quite a lot of stuff!

PR #2342, opened 2026-02-27
Fix meetup duplication

benjie:
Events (including locals) are pulled from the [GraphQL Foundation Calendar](https://calendar.graphql.org). The meetups file is just a list of locals, it is NOT a list of their events. This PR removes the duplication by stripping next/prev from locals so the only way to source upcoming/prior locals is via the calendar events. If events are missing on the foundation calendar, let me know.


jemgillam:
This preview looks good. At the moment it means all previous meetups have been removed from the "previous events" section, but they can presumably be put back by adding them to the google calendar. I like this proposed way of managing the events going forwards.

PR #2389, opened 2026-04-28
Update conference kit banners

hasparus:
This PR adds a `/conf/conference-kit` page we can use to iterate and download the roll-up banners, and maybe in the future stickers and other things.

I realized having this in the repo makes the most sense, because importing / copying existing text, design and icons is way faster and easier than trying to get something consistent and usable from an AI agent :)

I pulled in `qrcode` library to generate QR codes here.

image

Take note that we must run the `pnpm conference-kit` to make the zip archive, it doesn't get updated in normal website build.


martinbonnin (approved):
Thanks!

bignimbus (approved):
Amazing! Thank you!

PR #2401, opened 2026-05-07
Fix logo size of sponsor

benjie:
AirBnB's logo was rendering very small due to embedded padding. This PR renders logos with updated viewBox so that padding is removed to give logos equivalent sizes in the same tier. It also matches the rank of the silver and open source sponsor tiers, and improves padding between the logos and the tier division lines.

I've also noted why we couldn't do the same treatment to Meta.


martinbonnin:
FWIW, we've had the same issue in the app and [ended up rendering the sponsors in PNG](https://github.com/apollographql/graphqlconf-app/tree/main/kotlin/sponsor-logos)

PR #2427, opened 2026-06-08
Fix dead link: 'Learn more' button on community events page

aexol:
The 'Learn more' button in the 'Bring GraphQL to your community' section had `href="#"` with a TODO comment. Fixed it to point to `/community/foundation/local-initiative/` — the page that explains how to start a local GraphQL group.


benjie (approved):
These both point to the same place now. Maybe the start button should use an anchor tag to take you right to the bottom of the target page with the contact details? Either way this is better than what we have currently so I’m merging; thanks!

PR #2434, opened 2026-06-16
Add best practice about robust applications

martinbonnin:
Robust applications must:
  * Plan for unknown enum values
  * Plan for unknown union types being added
  * Not force unwrap nullable types


src/pages/learn/robust-applications.mdx
```diff
@@ -0,0 +1,121 @@
...
+  case "ACTIVE":
+    return showActive()
+  case "INACTIVE":
+    return showInactive()
+  default:
+    return showUnknown(status)
+}
+```
+
+In typed languages, if your codegen tool marks enums as exhaustive, configure it to generate a catch-all variant (often called `UNKNOWN` or `%future added value`) so the compiler enforces that you handle it.
```
PascalSenn, line 40: So - we should have an admonition here.. as `%future added value` is fake in some clients

mjmahone (approved):
This is great!

leebyron (approved):

PR #2440, opened 2026-06-21
Add PSGraphQL to language support

anthonyg-1:
## Description

Adds PSGraphQL as a PowerShell GraphQL client library under Language Support.

PSGraphQL is a PowerShell module for querying and mutating GraphQL endpoints from scripts, shells, and automation workflows. It is published on the PowerShell Gallery and supports Windows, Linux, and macOS.

Changes:
- Adds `PowerShell` to the Code page slug map
- Adds `PSGraphQL` under `language-support/powershell/client`

### DIFF
--- a/src/code/language-support/powershell/client/psgraphql.md
+++ b/src/code/language-support/powershell/client/psgraphql.md
@@ -1,0 +1,38 @@
+---
+
+name: PSGraphQL
+description: A PowerShell module for querying and mutating GraphQL endpoints.
+github: anthonyg-1/PSGraphQL
+tags:
+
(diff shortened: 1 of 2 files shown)


martinbonnin (approved):

martinbonnin:
@anthonyg-1 the build fails with `reason: 'name of an alias node must contain at least one character',`. Can you double check?

anthonyg-1:
> 

Thanks @martinbonnin, good catch. The Markdown frontmatter had malformed YAML for the `tags` list. I fixed the frontmatter and pushed an update in `7c55c59`.

PR #2442, opened 2026-06-23
Add "workshop" events

martinbonnin:
Add a new category of events: "workshops". Also add @graphqlguy workshop as the first workshop.

Ping @jemgillam @benjie @Urigo 

This is how it looks:

Screenshot 2026-06-23 at 14 20 10

### DIFF
--- a/src/app/(main)/community/events/event-card.tsx
+++ b/src/app/(main)/community/events/event-card.tsx
@@ -55,7 +55,7 @@ export interface EventCardProps {
   city: ReactNode
   name: ReactNode
   meta?: ReactNode
-  kind: "meetup" | "conference" | "working-group" | "foundation-meeting"
+  kind: EventKind
   className?: string
 }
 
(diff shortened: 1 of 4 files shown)


jemgillam:
Hmmm we need to think about this some more, things which are promoted on this page are either Foundation run events or working groups. Though we have before promoted meet ups which aren't part of the Locals program, these are free to attend. 

Arguably, GraphQL Day is a paid event, but is directly supported as a Foundation event

J on the Beach is a paid event, though they do have community tickets available

Do we promote only workshops ran by Ambassadors? 
Do we promote any workshop ran by a GraphQL vendor at a non-GraphQL event?

Urigo:
I favor opening it as wide as possible to start with

It would mean that page becomes the place to look for all these types of
events, including vendor events

martinbonnin:
Agree with @Urigo. 

Those third party events help GraphQL grow and would be great to have them listed somewhere. Also we don't have that many at the moment. If it becomes too noisy/spammy we can always revisit.

martinbonnin:
I've added this topic to the [primary working session on Monday agenda](https://docs.google.com/document/d/19-alP5jywnXzgN_1zYLBTRWh-4CaXzGakEZdTBFwNAc/edit?tab=t.0) to gather more input.

jemgillam:
So long as there is a clear guideline I am happy!

benjie:
Have discussed this with Lee, we agree with Uri that widening the net of this page would be great. The main caveat is it needs to be very obvious that these community or vendor hosted events are marked as such, and are clearly NOT hosted by GraphQL Foundation.

The initial suggestion is that non-Foundation events should have a bold "Hosted by [NAME]" callout. Would be great to add a number of GraphQL-focussed non-Foundation events right from the start: GraphQL Day ("Hosted by Team GraphQL Day"?), Apollo Summit ("Hosted by Apollo"), this workshop ("Hosted by Željko Kozina"), etc.

martinbonnin:
Notes from the grooming session: it's on me to suggest a design that makes it clear what events are 3rd parties and what events are not.

PR #2444, opened 2026-06-23
Fix the way that lightning talks render on the schedule

benjaie:
Really ought to have done this before the conference...


jemgillam (approved):

jemgillam:
I like that the lightning talks line up, but doing so has pushed the speaker line down in the other talks in the other columns

Screenshot 2026-06-25 at 09 27 05

jemgillam:
@benjie here is my suggestion for it to look a bit better: 

Screenshot 2026-06-25 at 10 57 38

I have pushed directly to the benjaie branch but it's just one commit which you could throw away if wanted.

jemgillam (approved):

PR #2462, opened 2026-07-02
Robust applications: add more language around tools with a runtime vs no-runtime

martinbonnin:
`%future added value` can be confusing so it's worth adding some extra language. Thanks @PascalSenn for raising this. Let me know how that looks. 

@eddeee888 the more I look into this, the more I think `%future added value` should more be like `%do not use this value, add a default case instead`

### DIFF
--- a/src/pages/learn/robust-applications.mdx
+++ b/src/pages/learn/robust-applications.mdx
@@ -37,7 +37,45 @@ switch (status) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


PascalSenn (approved):

PascalSenn:
@martinbonnin  +1 for `%do not use this value, add a default case instead`

eddeee888:
`%do not use this value, add a default case instead` sounds good  @martinbonnin @PascalSenn , we can build this default case in the next major

benjie:
@martinbonnin would be great if you added this advice to the golden path - helping users to build robust applications by having clients require that additional enums/object types are handled in code. A lint rule can also be added to ensure the special value never appears in user code - must be handled by the default case instead.

PR #2470, opened 2026-07-26
Add slides filter to conference schedule

### LINKED ISSUE #2420: Add support for filtering sessions that have slides attached/or not
This is a feature in sched. Would be nice to have the same here

Debagithub567:
Closes #2420

## Description

This PR adds support for filtering conference sessions based on whether presentation slides are attached.

### Changes

- Added a `hasSlides` property to `ScheduleSession`
- Derived the `hasSlides` value from the existing `session.files` data
- Added a **Slides** filter to the conference schedule
- Reused the existing generic filtering mechanism without introducing additional filtering logic

### Testing

- Verified that the **Slides** filter appears in the conference schedule
- Verified that sessions with attached files appear under **Yes**
- Verified that sessions without attached files appear under **No**
- Confirmed that the filter works alongside the existing schedule filters

### Screenshot

The new **Slides** filter with **Yes/No** options in the conference schedule.

graphql_slides_filter

### DIFF
--- a/src/app/conf/2026/_data.ts
+++ b/src/app/conf/2026/_data.ts
@@ -6,11 +6,13 @@ import { readSpeakers } from "../_api/sched-data"
 const speakersData = require("../../../../scripts/sync-sched/speakers.json")
 const equalitySets: string[][] = speakersData.equal || []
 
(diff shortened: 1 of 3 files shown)


martinbonnin (approved):
Thank you!

PR #2472, opened 2026-07-28
Fix formatting and improve clarity in schema.mdx

narsulet:
## Description

The text for the example for `null` argument is in the wrong place. Notice it says `As we saw in the example above`, even though the example it is referring to is below.


benjie (commented):
Thanks for your pull request; however this was correct as-is. The example that follows the colon is showing an invalid query, its referencing the non-null in the definition of the `Query.droid(id:)` argument:

```graphql
type Query {
  droid(id: ID!): Droid
}
```

There may be confusion here, but this is not the correct fix. Please feel free to raise a new PR that clarifies within the paragraph; but note that we will be unable to merge any fix until the CLA is signed.

PR #2477, opened 2026-08-01
Add Turbofy to GraphQL services

Jeansse:
## Description

Add Turbofy - an API-first vibe-coding platform with a GraphQL playground for their embedded GraphQL workspace API - to the GraphQL.org Tools & Libraries directory and the frontend, backend, and AI resource hubs.


martinbonnin (approved):
There's probably a larger discussion to be had about what exactly is a service but I'm overall agreeing to this change.

PR #2483, opened 2026-08-11
Fix conference planning event categorization

hasparus:
Treats conference planning sessions as working group meetings. Event filters keep one category visible, and `/events` redirects to `/community/events`.

Currently on prod, Conference planning calls render the same way as conferences, and while they're neither working groups nor foundation meetings, IMHO they're definitely not conferences and they shouldn't be sorted with them.

## Before

image


martinbonnin (approved):
🎉

PR #2494, opened 2026-08-26
Add Toronto meet up

jemgillam:
## Description

Add the new Toronto Local to the events list! Congrats @egoodwinx we look forward to hearing how it goes! (We are also here to support you with any help you may need.) 

The event itself has been added to the Foundation calendar, so should show up on the website when it next gets published. If you're in the Toronto area, why not come along to the first meet up on 7th October? https://guild.host/graphql-toronto 

*Also removes a meet up which never got off the ground*

### DIFF
--- a/src/components/meetups.ts
+++ b/src/components/meetups.ts
@@ -80,19 +80,6 @@ export const meetups = [
   },
   {
     node: {
-      id: "bialystock",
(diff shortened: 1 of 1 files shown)


martinbonnin (approved):

egoodwinx:
Awesome - thank you!

PR #2507, opened 2026-09-11
Update to GraphQL Days websites

jemgillam:
Add Bengaluru schedule, talk information, speaker photos etc

Make Singapore and NYC schedule visible

Remove CFP & ticket buttons where not needed


martinbonnin (approved):
Very nice 👍

